PR #21, opened 2023-10-26
Support dynamic refresh

Eskibear:
#### Design

New APIs:
```ts
interface AzureAppConfiguration {
  ...
  refresh(): Promise<void>; // api to trigger refresh
  onRefresh(listener: () => any, thisArg?: any): void; // success callback
}

interface AzureAppConfigurationOptions { 
    ... 
    refreshOptions?: RefreshOptions; 
}

interface RefreshOptions {
  enabled: boolean; // default to false
  refreshIntervalInMs?: number; // default to 10s
  watchedSettings?: WatchedSetting[]
}

interface WatchedSetting  { 
    key: string; 
    label?: string;
}
```

Refresh Behavior:
image


src/AzureAppConfigurationImpl.ts
```diff
@@ -29,20 +35,32 @@ export class AzureAppConfigurationImpl extends Map<string, unknown> implements A
...
         this.requestTracingEnabled = requestTracingEnabled();
-        if (this.requestTracingEnabled) {
-            this.enableRequestTracing();
-        }
 
         if (options?.trimKeyPrefixes) {
             this.sortedTrimKeyPrefixes = [...options.trimKeyPrefixes].sort((a, b) => b.localeCompare(a));
         }
+
+        if (options?.refreshOptions) {
```
Eskibear, line 43: TODO: `watchedSettings` doesn't support wildcard, we should validate somewhere, and throw error if `*` detected.

src/WatchedSetting.ts
```diff
@@ -0,0 +1,7 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT license.
+
+export interface WatchedSetting {
+    key: string;
+    label?: string;
```
Eskibear, line 6: JS SDK:
In client.listConfigurationSettings(options) , it accepts labelFilter as `\0` for "no label".
image

while in the response, "no label" represented by `null`:
image

When not specified, e.g. `{key: "some-key"}`, then the label is `undefined`. 

Here we must consider how to deal with these values and well-document them:
- `\0`
- `null`
- `undefined`

Eskibear: image

Here by omitting it, it means "no label" instead of "any" as above in LabelFilter... Ideally I would expect something to be done in upstream JS SDK, to eliminate the inconsistency...

Eskibear: created a ticket in upstream to track: https://github.com/Azure/azure-sdk-for-js/issues/27607

src/AzureAppConfigurationImpl.ts
```diff
@@ -29,20 +36,41 @@ export class AzureAppConfigurationImpl extends Map<string, unknown> implements A
         super();
         // Enable request tracing if not opt-out
         this.requestTracingEnabled = requestTracingEnabled();
-        if (this.requestTracingEnabled) {
-            this.enableRequestTracing();
-        }
 
         if (options?.trimKeyPrefixes) {
             this.sortedTrimKeyPrefixes = [...options.trimKeyPrefixes].sort((a, b) => b.localeCompare(a));
```
juniwang, line 41: why do we need to `sort`>

src/AzureAppConfigurationImpl.ts
```diff
@@ -29,20 +36,41 @@ export class AzureAppConfigurationImpl extends Map<string, unknown> implements A
...
+            const refreshIntervalInMs = this.options?.refreshOptions?.refreshIntervalInMs;
+            if (refreshIntervalInMs !== undefined) {
+                if (refreshIntervalInMs < MinimumRefreshIntervalInMs) {
+                    throw new Error(`The refresh interval time cannot be less than ${MinimumRefreshIntervalInMs} milliseconds.`);
+                } else {
+                    this.refreshIntervalInMs = refreshIntervalInMs;
+                }
+            }
+
+            this.sentinels = options.refreshOptions.watchedSettings?.map(setting => {
```
juniwang, line 57: why it's different between `options.refreshOptions.watchedSettings` and `this.options?.refreshOptions?.refreshIntervalInMs`? where the latter calls `this` and uses nullable`?.`

src/AzureAppConfigurationImpl.ts
```diff
@@ -55,23 +83,80 @@ export class AzureAppConfigurationImpl extends Map<string, unknown> implements A
...
         }
         for (const [k, v] of keyValues) {
             this.set(k, v);
         }
+        this.lastUpdateTimestamp = Date.now();
+    }
+
+    public async refresh(): Promise<void> {
+        // if no refreshOptions set, return
+        if (this.sentinels === undefined || this.sentinels.length === 0 || this.refreshIntervalInMs === undefined) {
```
juniwang, line 112: no default value for `refreshIntervalInMs`?

src/AzureAppConfigurationImpl.ts
Eskibear: The purpose is to trim the longest prefix first, if one is substring of another. E.g. 
prefixes = ["app.", "app.settings."], we expect "app.settings.message" to be trimmed as "message".

src/AzureAppConfigurationImpl.ts
Eskibear: default value is 30s. Currently it's assigned in the constructor if refreshOptions is specified. Otherwise it's undefined.

juniwang: And to confirm, we are supporting collection monitoring for now, at least not included in this PR. Is this true?

src/AzureAppConfigurationImpl.ts
Eskibear: 1. `options` is equivalent to `this.options` in the constructor. It's a grammar sugar.
```ts
// simplified
class Foo {
  constructor(private bar: string) {}
}

// equivalent
class Foo {
  private bar: string;
  constructor(bar: string) {
    this.bar = bar;
  }
}
```

src/AzureAppConfigurationImpl.ts
```diff
@@ -55,23 +83,80 @@ export class AzureAppConfigurationImpl extends Map<string, unknown> implements A
...
+                    listener();
+                }
+            }
+        }
+    }
+
+    public onRefresh(listener: () => any, thisArg?: any): Disposable {
+        if (this.onRefreshListeners === undefined) {
+            // TODO: Add unit tests
+            throw new Error("Illegal operation because refreshOptions is not provided on loading.");
```
juniwang, line 149: I see `onRefreshListeners` is initialized in contructor. Is there a code path that it's not initialized ?

src/AzureAppConfigurationImpl.ts
Eskibear: 2. about `?.`, it's another convenient grammar sugar. 

```ts
// short
a?.b 
// equivalent
a ? a.b : undefined
```

Without `?` it's possible to have NPE like access some property of undefined. The chained `?.` will be super convenient if you want to access something deep inside optional properties.

src/AzureAppConfigurationImpl.ts
Eskibear: It's initialized as a linkedList only when refreshOptions is specified. So if no refreshOptions provided, it's undefined. 

An alternative impl is we always initialize it as an empty list, if we want to get rid of the null checks.

src/AzureAppConfigurationImpl.ts
Eskibear: In this PR, the supported refresh is sentinel-based. You pass `watchedSettings` as a list of sentinels, the provider periodically calls getConfigurationSetting api to check if any of them is modified. If modification is detected, the provider then re-load all the configuration settings to update the Map.

src/WatchedSetting.ts
Eskibear: azure-app-configuration v1.5.0 released containing the fix. we should update.

Eskibear: update with PR https://github.com/Azure/AppConfiguration-JavaScriptProvider/pull/27

src/AzureAppConfigurationImpl.ts
juniwang: I mean what's difference between `watchedSettings` and `refreshIntervalInMs`? one needs null check, another doesn't

Eskibear: Good catch.  They are in the block that `options.refreshOptions` is not null. But `this.options.refreshOptions` can be null (impossible, only from perspective of TypeScript type deduction), and that's why I added the null check.

I shall change `this.options?.refreshOptions?.refreshIntervalInMs` to `options.refreshOptions.refreshIntervalInMs`.

And based on the discussion we'll add a property `enabled` in `refreshOptions`, explictly indicating whether to enable the feature. I'll re-organize and simplify the impl later.

src/WatchedSetting.ts
Eskibear: fixed by upstream sdk.

src/AzureAppConfigurationImpl.ts
Eskibear: I will always init refreshListener as an emtpy list. And use refreshOptions.enabled to determine whether refresh is enabled and the onRefresh API is available.

src/AzureAppConfigurationImpl.ts
Eskibear: checked and added a test case. Now refreshOptions.watchedSettings will be validated and saved into a list `_sentinels`.

src/AzureAppConfigurationImpl.ts
Eskibear: fixed.

src/AzureAppConfigurationImpl.ts
Eskibear: updated.

src/refresh/RefreshTimer.ts
```diff
@@ -0,0 +1,61 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT license.
+
+const MinimumBackoffInMs = 30 * 1000; // 30s
+const MaximumBackoffInMs = 10 * 60 * 1000; // 10min
+const MaxAttempts = 63;
+const JitterRatio = 0.25;
+
+export class RefreshTimer {
```
Eskibear, line 9: @mrm9084 @avanigupta please take a look as I took both Python and .NET for reference when implementing the backoff part.

src/AzureAppConfigurationImpl.ts
```diff
@@ -29,20 +37,49 @@ export class AzureAppConfigurationImpl extends Map<string, unknown> implements A
...
+                if (refreshIntervalInMs < MinimumRefreshIntervalInMs) {
+                    throw new Error(`The refresh interval time cannot be less than ${MinimumRefreshIntervalInMs} milliseconds.`);
+                } else {
+                    this.refreshInterval = refreshIntervalInMs;
+                }
+            }
+
+            this.sentinels = watchedSettings.map(setting => {
+                if (setting.key.includes("*") || setting.label?.includes("*")) {
+                    throw new Error("The character '*' is not supported in key or label.");
```
RichardChen820, line 63: Watch settings with wildcard key or label is not supported

Eskibear: I was following python's impl checking wildcard char `*`. In .NET there's no such check for watched settings. Characters `*`  and `,` are just regarded as plain text, and thus won't match any valid setting in most of time.

So now I'm thinking whether we need the check.
- Option a) keep the check. Then we probably also want to add `,` as well.
- Option b) remove the check. Then we need to document or add doc comments to the API.

src/AzureAppConfigurationImpl.ts
```diff
@@ -74,6 +117,68 @@ export class AzureAppConfigurationImpl extends Map<string, unknown> implements A
...
+
+        // if still within refresh interval/backoff, return
+        if (this.refreshTimer.isDuringBackoff()) {
+            return Promise.resolve();
+        }
+
+        // try refresh if any of watched settings is changed.
+        let needRefresh = false;
+        for (const sentinel of this.sentinels) {
+            const response = await this.client.getConfigurationSetting(sentinel, {
```
RichardChen820, line 136: What if any error happen while querying the sentinel key? Seems we should handle it as well.

Eskibear: I agree that for RestError with 404 we should handle it, as it's a valid case when the setting is deleted.

For other unexpected errors, shall we silent fail in order not to break the app?

src/AzureAppConfigurationImpl.ts
```diff
@@ -29,20 +37,49 @@ export class AzureAppConfigurationImpl extends Map<string, unknown> implements A
...
+            const { watchedSettings, refreshIntervalInMs } = options.refreshOptions;
+            // validate watched settings
+            if (watchedSettings === undefined || watchedSettings.length === 0) {
+                throw new Error("Refresh is enabled but no watched settings are specified.");
+            }
+
+            // custom refresh interval
+            if (refreshIntervalInMs !== undefined) {
+                if (refreshIntervalInMs < MinimumRefreshIntervalInMs) {
+                    throw new Error(`The refresh interval time cannot be less than ${MinimumRefreshIntervalInMs} milliseconds.`);
```
avanigupta, line 55:
```suggestion
                    throw new Error(`The refresh interval cannot be less than ${MinimumRefreshIntervalInMs} milliseconds.`);

```

src/AzureAppConfigurationImpl.ts
```diff
@@ -55,17 +92,23 @@ export class AzureAppConfigurationImpl extends Map<string, unknown> implements A
...
-                    const trimmedKey = this.keyWithPrefixesTrimmed(key);
-                    keyValues.push([trimmedKey, value]);
+                    const keyValuePair = await this.processKeyValues(setting);
+                    keyValues.push(keyValuePair);
+                }
+                // update etag of sentinels if refresh is enabled
+                if (this.refreshEnabled) {
+                    const matchedSentinel = this.sentinels.find(s => s.key === setting.key && s.label === setting.label);
+                    if (matchedSentinel) {
+                        matchedSentinel.etag = setting.etag;
```
avanigupta, line 110: Just confirming if this update to `matchedSentinel` will also update the corresponding etag in `this.sentinels`?

src/AzureAppConfigurationImpl.ts
```diff
@@ -74,6 +117,68 @@ export class AzureAppConfigurationImpl extends Map<string, unknown> implements A
...
+    /**
+     * Refresh the configuration store.
+     */
+    public async refresh(): Promise<void> {
+        if (!this.refreshEnabled) {
+            return Promise.resolve();
+        }
+
+        // if still within refresh interval/backoff, return
+        if (this.refreshTimer.isDuringBackoff()) {
```
avanigupta, line 129: How about naming the method something like `canRefresh()`? That covers both backoff and normal refresh interval wait period. Then we can check `if (!this.refreshTimer.canRefresh())`

src/refresh/RefreshTimer.ts
```diff
@@ -0,0 +1,64 @@
...
+            minBackoffMs = MinimumBackoffInMs
+            maxBackoffMs = this._interval;
+        } else {
+            minBackoffMs = MinimumBackoffInMs;
+            maxBackoffMs = MaximumBackoffInMs;
+        }
+
+        // exponential
+        let calculatedBackoffMs = Math.max(1, minBackoffMs) * (1 << Math.min(this._attempts, MaxAttempts));
+        if (calculatedBackoffMs > maxBackoffMs) {
```
avanigupta, line 55: Is there any risk of overflow for `calculatedBackoffMs`?

src/AzureAppConfigurationImpl.ts
avanigupta: We should bubble up all errors and increase the backoff time. 
In .NET, this immediately breaks the app if `RefreshAsync` is invoked. But if users are invoking `TryRefreshAsync`, app will not crash on known errors (like `RequestFailedException`, `AggregateException`, etc). App can still crash on unknown errors.
In Python, we provide an additional error callback that can be invoked if there are any known errors. If no callback is provided, or an unknown error occurs, app will crash.

src/AzureAppConfigurationImpl.ts
avanigupta: Now that we are planning collection monitoring support, I think it's a good time to consider if we want to block these characters at all. Let's sync offline about this.

Eskibear: The idea is `watchedSettings` takes a list of specific configuration setting ids. To watch wildcard patterns I think we should resort to collection monitoring. 
In this PR I will take option a) to block wildcard chars like `*` and `,`.  And in the error message I will highlight "watched settings" and specific chars, as suggested.

src/AzureAppConfigurationImpl.ts
RichardChen820: We didn't design to have an API `TryRefresh()` here right? Probably we could do similar with Python, provide an error callback that can be invoked if any known error happen

src/refresh/RefreshTimer.ts
Eskibear: The overflow is not as critical as in other language.  Numbers (double) larger than [Number.MAX_VALUE](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Number/MAX_VALUE) is `infinity` which is larger than any number in comparison.

And [Number.MAX_SAFE_INTEGER](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Number/MAX_SAFE_INTEGER) is the largest **integer** that ensures comparison to work correctly.

for simplicity, current logic to check `calculatedBackoffMs > maxBackoffMs` should work even when overflow occurs, AFAIK. 

But yes, it's better to add overflow guardian, using MAX_SAFE_INTEGER. I'll update accordingly.

src/AzureAppConfigurationImpl.ts
Eskibear: `canRefresh` looks good. 

BTW I used to refer to `needs_refresh` from python, which is a little bit unclear to me, because it doesn't really "need" a refresh, it stands for whether a real refresh is allowed.

src/AzureAppConfigurationImpl.ts
Eskibear: Yes, it will. `find` returns the reference of the object. By modifying its properties, it affects this.sentinels which actually is a list of references.

src/common/linkedList.ts
```diff
@@ -0,0 +1,140 @@
...
+	prev: Node<E>;
+
+	constructor(element: E) {
+		this.element = element;
+		this.next = Node.Undefined;
+		this.prev = Node.Undefined;
+	}
+}
+
+export class LinkedList<E> {
```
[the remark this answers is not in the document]
Eskibear: It's copied from vscode's repo. The main purpose was to get a callback which can easily remove the element when calling push/insert, see Line 78-83. And the performance boost is by the way, nice to have.

Using array, after I insert an element, I have to keep the element and perform a linear search, and slice the array. It's feasible. But that's a little bit more complicated, comparing to above existing solution.

src/RefreshOptions.ts
```diff
@@ -0,0 +1,27 @@
...
+import { WatchedSetting } from "./WatchedSetting";
+
+export const DefaultRefreshIntervalInMs = 30 * 1000;
+export const MinimumRefreshIntervalInMs = 1 * 1000;
+
+export interface RefreshOptions {
+    /**
+     * Specifies whether the provider should automatically refresh when the configuration is changed.
+     */
+    enabled: boolean;
```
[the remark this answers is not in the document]
Eskibear: per design review, it's for explicitness.

src/AzureAppConfigurationImpl.ts
Eskibear: @RichardChen820 
No, we don't provide `TryRefresh()`, as user can wrap `refresh()` with try-catch blocks on their own, and deal with different kinds of errors.

JS is a language full of callbacks. The `refresh()` returns a promise, which works with callbacks naturally.
```js
Promise.then(successCallback).catch(errorCallback);
```
Note that, we do have an dedicated `onRefresh()` API to pass success callbacks, serving for the purpose that you want to set callbacks in elsewhere, other than the place you call refresh(). Maybe if necessary, later we can have an onRefreshFailure() API for error callback...

src/AzureAppConfigurationImpl.ts
Eskibear: updated in https://github.com/Azure/AppConfiguration-JavaScriptProvider/pull/21/commits/1da9647555fbbf19d049d10a02ed5f18a7ea513e

src/common/linkedList.ts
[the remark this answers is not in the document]
Eskibear: updated in e72887a, removing linked list and use array instead.

src/AzureAppConfigurationImpl.ts
Eskibear: updated in 5b3ab06

examples/refresh.mjs
```diff
@@ -0,0 +1,44 @@
...
+import * as dotenv from "dotenv";
+import { promisify } from "util";
+dotenv.config();
+const sleepInMs = promisify(setTimeout);
+
+/**
+ * This example retrives all settings with key following pattern "app.settings.*", i.e. starting with "app.settings.".
+ * With the option `trimKeyPrefixes`, it trims the prefix "app.settings." from keys for simplicity.
+ * Value of config "app.settings.message" will be printed.
+ *
```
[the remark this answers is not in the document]
Eskibear: updated in 18bbd4e

src/refresh/RefreshTimer.ts
Eskibear: Just revisit the code, if "overflow" occurs and calculatedBackoffMs  > MAX_SAFE_INTEGER, it is certainly larger than maxBackoffMs.
So, no change is needed.

src/AzureAppConfigurationImpl.ts
avanigupta: With that Promise pattern, I think we can safely throw the errors and have user code handle it. The only one we need to catch is 404 like you said, and make sure to increment backoff time on any error.

src/common/linkedList.ts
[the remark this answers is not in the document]
avanigupta: My concern was about maintainability of custom data structure code. If there's not much difference in performance and code complexity, I would prefer to use built in types like arrays. The new implementation looks good to me, but if you feel arrays are increasing the complexity or significantly reducing performance, feel free to use LinkedList.

src/AzureAppConfigurationImpl.ts
```diff
@@ -79,6 +127,75 @@ export class AzureAppConfigurationImpl extends Map<string, unknown> implements A
...
+    /**
+     * Refresh the configuration store.
+     */
+    public async refresh(): Promise<void> {
+        if (!this.#refreshEnabled) {
+            return Promise.resolve();
+        }
+
+        // if still within refresh interval/backoff, return
+        if (this.#refreshTimer.canRefresh()) {
```
avanigupta, line 139:
```suggestion
        if (!this.#refreshTimer.canRefresh()) {

```

src/AzureAppConfigurationImpl.ts
avanigupta: In Python, the variable `need_refresh` is used to track if any sentinel keys have changed on the server. You have a similar variable which makes sense because we are tracking if configuration needs to be refreshed after a server call has already been made.

src/AzureAppConfigurationOptions.ts
```diff
@@ -35,4 +36,8 @@ export interface AzureAppConfigurationOptions {
      * Specifies options used to resolve Vey Vault references.
      */
     keyVaultOptions?: AzureAppConfigurationKeyVaultOptions;
+    /**
+     * Specifies options for dynamic refresh key-values.
+     */
+    refreshOptions?: RefreshOptions;
```
avanigupta, line 42: I know we finalized the name `RefreshOptions` because `AzureAppConfigurationOptions` already contains the prefix `AzureAppConfiguration`. 
Looking at other names in this class, I'm wondering if we should change those to drop the app config prefix too?
- `AzureAppConfigurationKeyVaultOptions` -> `KeyVaultOptions`
- `AppConfigurationClientOptions` -> `ConfigurationClientOptions`

src/common/linkedList.ts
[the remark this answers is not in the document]
Eskibear: OK. Let's just keep it as is, using the array first. (I'd rather not to maintain the data structure...) And we can switch to linked list if performance degradation and customer impact observed.

src/AzureAppConfigurationImpl.ts
Eskibear: by bad. wondering why the test case didn't cover it. Let me check.

src/AzureAppConfigurationOptions.ts
Eskibear: Good point. I agree.

Eskibear: To make it clear, I'll create a separate PR updating it later.

src/AzureAppConfigurationImpl.ts
Eskibear: Got it. I didn't change the impl either.
```
    public canRefresh(): boolean {
        return Date.now() < this._backoffEnd;  // should be >=
    }
```

Eskibear: fixed in 0e347be

src/AzureAppConfigurationOptions.ts
Eskibear: created #44

src/AzureAppConfigurationImpl.ts
Eskibear: I meant [RefreshTimer.needs_refresh](https://github.com/Azure/azure-sdk-for-python/blob/98274a8e4684d868471654cfe92469d452316428/sdk/appconfiguration/azure-appconfiguration-provider/azure/appconfiguration/provider/_azureappconfigurationprovider.py#L408-L410).

src/common/disposable.ts
```diff
@@ -0,0 +1,15 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT license.
+
+export class Disposable {
```
zhiyuanliang-ms, line 4: Should this be put in the index.ts?

src/refresh/RefreshTimer.ts
```diff
@@ -0,0 +1,64 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT license.
+
+const MinimumBackoffInMs = 30 * 1000; // 30s
+const MaximumBackoffInMs = 10 * 60 * 1000; // 10min
+const MaxAttempts = 63;
```
zhiyuanliang-ms, line 6: Can we add some comments to explain these numbers?

src/AzureAppConfigurationImpl.ts
Eskibear: 404 handled in 46be338 .

src/common/disposable.ts
Eskibear: exactly. good catch!

src/refresh/RefreshTimer.ts
```diff
@@ -0,0 +1,80 @@
...
+    private _minBackoff: number;
+    private _maxBackoff: number;
+    private _attempts: number;
+    private _backoffEnd: number; // Timestamp
+    constructor(
+        private _interval: number
+    ) {
+        this._minBackoff = Math.min(this._interval, MinimumBackoffInMs);
+        this._maxBackoff = Math.min(this._interval, MaximumBackoffInMs);
+        this._attempts = 0;
```
zhiyuanliang-ms, line 34: In python provider, _attempts is initialized as 1.

src/refresh/RefreshTimer.ts
```diff
@@ -0,0 +1,80 @@
...
+const MaxAttempts = 63;
+const JitterRatio = 0.25;
+
+export class RefreshTimer {
+    private _minBackoff: number;
+    private _maxBackoff: number;
+    private _attempts: number;
+    private _backoffEnd: number; // Timestamp
+    constructor(
+        private _interval: number
```
zhiyuanliang-ms, line 30: What will happen if _interval is negative?

Eskibear: Interval cannot be set to negative. Its min value is 1000ms.

zhiyuanliang-ms: It seems like the python provider lacks the check for invalid refresh interval. I will talk to Matthew offline.

Eskibear: But I agree that in terms of the RefreshTimer itself, we should check the interval passed into constructor and ensure it's positive.

src/refresh/RefreshTimer.ts
Eskibear: AFAIU, the `MaxAttempt` (copied from dotnet provider) should be the guardian of overflow for 64 bit integer. 
Here Number.MAX_SAFE_INTEGER = 2^53-1, so in JS we should use 53. 
And the name is confusing. I would use `MaxSafeExponential` instead.

src/refresh/RefreshTimer.ts
Eskibear: fixed in e0c9736

src/refresh/RefreshTimer.ts
Eskibear: fixed as above in e0c9736

src/refresh/RefreshTimer.ts
Eskibear: more details about the timer added in ead82d0

src/common/disposable.ts
Eskibear: fixed in 7663a27

src/refresh/RefreshTimer.ts
```diff
@@ -0,0 +1,106 @@
...
+    ) {
+        if (this._interval <= 0) {
```
zhiyuanliang-ms, line 32: The `Math.min()` here can be removed. In `_CalculateBackoffTime()`, we have `if (this._interval <= this._minBackoff)`

src/refresh/RefreshTimer.ts
zhiyuanliang-ms: Discussed with Yan offline. `_attempts` will be renamed to `_failedAttempts`

src/refresh/RefreshTimer.ts
Eskibear: Good catch. It's duplicate.

Since the logic is one-time initialization given fixed interval and min/max backoff time, I think it's better to move this part into constructor instead of in `_calculateBackoffTime`. It avoids repeated calculations.

Eskibear: Let me first remove the duplicate code from constructor. let's see how it looks like.

src/refresh/RefreshTimer.ts
```diff
@@ -0,0 +1,103 @@
...
+ *  - If the interval is between the minimum and maximum backoff, the interval is used as the maximum backoff.
+ *  - Because of the jitter, the maximum backoff time is actually `MaximumBackoffInMs` * (1 + `JitterRatio`).
+ */
+
+const MinimumBackoffInMs = 30 * 1000; // 30s
+const MaximumBackoffInMs = 10 * 60 * 1000; // 10min
+const MaxSafeExponential = 53; // Used to avoid overflow, as Number.MAX_SAFE_INTEGER = 2^53 - 1.
+const JitterRatio = 0.25;
+
+export class RefreshTimer {
```
zhiyuanliang-ms, line 24: I think the class name RefreshTimer is misleading. I think backoffTimer is more appropriate.

src/RefreshOptions.ts
```diff
@@ -0,0 +1,27 @@
...
+
+    /**
+     * Specifies the minimum time that must elapse before checking the server for any new changes.
+     * Default value is 30 seconds. Must be greater than 1 second.
+     * Any refresh operation triggered will not update the value for a key until after the interval.
+     */
+    refreshIntervalInMs?: number;
```
zhiyuanliang-ms, line 20: refreshInterval is misleading. This is actually for backoff.

zhiyuanliang-ms: try {
      await this.load(RequestType.Watch);
      this.#refreshTimer.reset();
  } catch (error) {
      // if refresh failed, backoff
      this.#refreshTimer.backoff();
      throw error;
  }
  If `load()` successfully each time we call `refresh()`, the refreshInterval will not work at all.

zhiyuanliang-ms: The refresh option in dotnet provider provide a method to set cache expiration interval for watched key.
https://github.com/Azure/AppConfiguration-DotnetProvider/blob/main/src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationRefreshOptions.cs#L58

The cache expiration time of watched key will be update through this method
https://github.com/Azure/AppConfiguration-DotnetProvider/blob/main/src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationProvider.cs#L941

@avanigupta

zhiyuanliang-ms:
Should the `refresh()` be thread-safe? Should we make sure the refresh() method can only be called by one thread?
Something like: https://github.com/Azure/AppConfiguration-DotnetProvider/blob/main/src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationProvider.cs#L175

src/RefreshOptions.ts
zhiyuanliang-ms: The python provider also uses this misleading name. @mrm9084

Eskibear: It’s affected by the exponential backoff, indeed we should update the comments here to clarify the behavior. I myself think CacheExpires is more confusing, because here actually no cache is preserved. What it does is, it prevents sending requests within a specific period of time, as if there is a cache. So that’s why we use the name “refresh interval” instead.

The refresh interval works when there is no failure. In the first few lines of refresh(), it checks #refreshTimer.canRefresh() , which checks if refresh interval is passed.

src/refresh/RefreshTimer.ts
Eskibear: I agree, to some extent, if we want to re-use the timer elsewhere other than refresh. Then, the timer is neutral, and the refresh functionality uses a "backoffTimer" to calculate the backoff time. And we might also rename the method "canRefresh" back to "isBackoffEnd"... ;) 

@avanigupta @mrm9084 WDYT?

RichardChen820:
> Should we make sure that the `refresh()` method can only be executed once at a time? Something like: https://github.com/Azure/AppConfiguration-DotnetProvider/blob/main/src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationProvider.cs#L175 Or https://github.com/Azure/azure-sdk-for-python/blob/main/sdk/appconfiguration/azure-appconfiguration-provider/azure/appconfiguration/provider/_azureappconfigurationprovider.py#L480
> 
> Another question is that: Should `refresh()` method be async?

JS is single-thread model, I don't think we should worry about it. Correct me if I was wrong

Eskibear:
Basically, JavaScript is single-threaded (except [Web Workers](https://developer.mozilla.org/en-US/docs/Web/API/Web_Workers_API/Using_web_workers) or [Worker Threads](https://nodejs.org/api/worker_threads.html) And the package is not expected to run on workers doing message communication...
So, no worry.

src/refresh/RefreshTimer.ts
RichardChen820: `isBackoffDue` sounds better.

src/refresh/RefreshTimer.ts
avanigupta: > In python provider, _attempts is initialized as 1.  Also checked the dotnet provider
> It seems like the python provider is wrong.

In .NET, we increase the number of attempts before calculating backoff: https://github.com/Azure/AppConfiguration-DotnetProvider/blob/e9a3aa5b49958bd6190fb91c7a16d10ac46e3967/src/Microsoft.Extensions.Configuration.AzureAppConfiguration/ConfigurationClientManager.cs#L84C1-L85

In Python, we calculate the backoff and then increase the attempts: https://github.com/Azure/azure-sdk-for-python/blob/98274a8e4684d868471654cfe92469d452316428/sdk/appconfiguration/azure-appconfiguration-provider/azure/appconfiguration/provider/_azureappconfigurationprovider.py#L404-L406

Both are correct, just doing things differently.

mrm9084: Python defaults to 1 attempt, not 0, so we increment after. My logic or it is that if you ever request the attempt number it is always the attempt you are on.

Eskibear: https://github.com/Azure/azure-sdk-for-python/blob/98274a8e4684d868471654cfe92469d452316428/sdk/appconfiguration/azure-appconfiguration-provider/azure/appconfiguration/provider/_azureappconfigurationprovider.py#L421

For python, if the initial value is 1, above calculated time is 2x min duration, which is not consistent with dotnet and JS in this PR. Might change it to attempts-1?

src/RefreshOptions.ts
avanigupta: The "cache expiration" naming in .NET has caused confusion in the past and we are trying to move away from it ([tracking work item](https://github.com/Azure/AppConfiguration-DotnetProvider/issues/350)).
The naming `refreshIntervalInMs` is appropriate here because we are asking users to set the minimum time between server calls to check for changes. So this is the interval between 2 refresh operations.
Backoff is our internal concept that comes into play when we fail to read key-values from the server. Users shouldn't need to know the implementation details of our error handling strategies. So "backoff" would not be a good name to add to any user facing properties.

src/refresh/RefreshTimer.ts
avanigupta: When a refresh operation succeeds, the next server refresh will be triggered only after the specified refresh interval has expired. During this time, the provider cannot initiate any server operation - not because it's in backoff state, but because its refresh interval has not elapsed. Provider will be in backoff state only when the previous server refresh has failed. In general, "backoff" indicates that something bad has happened. That's why I'm hesitant to use the term "backoff" here. Backoff is our implementation detail.

Eskibear: Ah yes, this whatever timer is coupled deeply with the refresh interval, and its purpose is to calculate the real interval after applying exponential backoff strategy. Then probably no need to reuse it, at least for the moment.

src/refresh/RefreshTimer.ts
avanigupta: The first time we calculate backoff, attempts should be 1. This is consistent in .NET and Python. Am I missing something?

Eskibear: https://github.com/Azure/AppConfiguration-DotnetProvider/blob/e9a3aa5b49958bd6190fb91c7a16d10ac46e3967/src/Microsoft.Extensions.Configuration.AzureAppConfiguration/Extensions/TimeSpanExtensions.cs#L68

If I understand correctly, in dotnet the attempts is 1, and the calculated time is min duration. In python, the attempts is also 1, but the calculated time is 2 * min duration.

avanigupta: Ah yes, you're right. I forgot about that extra check in .NET when attempts=1. Python should also do that @mrm9084

src/RefreshOptions.ts
Eskibear: Ok, then let's make no change here as explained.

src/AzureAppConfigurationImpl.ts
```diff
@@ -34,20 +42,53 @@ export class AzureAppConfigurationImpl extends Map<string, unknown> implements A
...
         this.#requestTracingEnabled = requestTracingEnabled();
-        if (this.#requestTracingEnabled) {
-            this.#enableRequestTracing();
-        }
 
         if (options?.trimKeyPrefixes) {
             this.#sortedTrimKeyPrefixes = [...options.trimKeyPrefixes].sort((a, b) => b.localeCompare(a));
         }
+
+        if (options?.refreshOptions?.enabled) {
```
avanigupta, line 50: Can we reuse `refreshEnabled()` method here?

src/AzureAppConfigurationImpl.ts
```diff
@@ -60,25 +101,111 @@ export class AzureAppConfigurationImpl extends Map<string, unknown> implements A
...
             this.set(k, v);
         }
     }
 
+    /**
+     * Refresh the configuration store.
+     */
+    public async refresh(): Promise<void> {
+        if (!this.#refreshEnabled) {
+            return Promise.resolve();
```
avanigupta, line 135: If someone tries to invoke `refresh()` without setting `refreshOptions.enabled=true`, should we throw an error instead of silently retuning?

src/AzureAppConfigurationImpl.ts
```diff
@@ -60,25 +101,111 @@ export class AzureAppConfigurationImpl extends Map<string, unknown> implements A
...
             for await (const setting of settings) {
                 if (setting.key) {
-                    const [key, value] = await this.#processAdapters(setting);
-                    const trimmedKey = this.#keyWithPrefixesTrimmed(key);
-                    keyValues.push([trimmedKey, value]);
+                    const keyValuePair = await this.#processKeyValues(setting);
+                    keyValues.push(keyValuePair);
+                }
+                // update etag of sentinels if refresh is enabled
+                if (this.#refreshEnabled) {
```
avanigupta, line 116: If a watched setting is not loaded as part of the kv selectors, how do we update its etag in `this.sentinels`?

src/AzureAppConfigurationImpl.ts
```diff
@@ -60,25 +101,111 @@ export class AzureAppConfigurationImpl extends Map<string, unknown> implements A
...
+        if (!this.#refreshTimer.canRefresh()) {
+            return Promise.resolve();
+        }
+
+        // try refresh if any of watched settings is changed.
+        let needRefresh = false;
+        for (const sentinel of this.#sentinels) {
+            let response: GetConfigurationSettingResponse | undefined;
+            try {
+                response = await this.#client.getConfigurationSetting(sentinel, {
```
avanigupta, line 148: If a sentinel key was not loaded initially, its etag will be null. Will that possibly throw an error when calling `getConfigurationSetting` with `onlyIfChanged` set to true?

src/refresh/RefreshTimer.ts
```diff
@@ -0,0 +1,103 @@
...
+        this._backoffEnd = Date.now() + this._interval;
+    }
+
+    public canRefresh(): boolean {
+        return Date.now() >= this._backoffEnd;
+    }
+
+    public backoff(): void {
+        this._backoffEnd = Date.now() + this._calculateBackoffTime();
+        this._failedAttempts += 1;
```
avanigupta, line 45: If we initialized `_failedAttempts` to 0, we should increment this before calculating the backoff time.

src/refresh/RefreshTimer.ts
```diff
@@ -0,0 +1,103 @@
...
+ * `1 << exp` is not used because it returns wrong results when exp >= 31.
+ */
+function efficientPowerOfTwo(positiveExponential: number) {
+    if (positiveExponential < 0) {
+        throw new Error("exponential must be a non-negative integer.");
+    } else if (positiveExponential > MaxSafeExponential) {
+        throw new Error(`exponential must be less than or equal to ${MaxSafeExponential}.`);
+    }
+
+    // bitwise operations in JavaScript are limited to 32 bits. It overflows at 2^31 - 1.
```
avanigupta, line 97: What do you think about changing the `MaxSafeExponential` to 30 then? In reality, we wouldn't reach that high number of attempts anyway.

avanigupta: The max backoff will always be less than 10 minutes. That means, `minBackoffMs * efficientPowerOfTwo` will always be capped at 600,000 ms (10 minutes). `2^20` already goes much higher than 600k, so we know that we would only backoff for 10 minutes after 19 failed attempts (max).

Eskibear: The number was determined by the language’s limitation, for safe operations. I don’t want it to be determined by the max backoff time, otherwise they are coupled too much and it’s risky to change max backoff time(just in case).

But yes, 30 is reasonable, as it’s determined by the bitwise operation limitation. It’s large enough for reasonable backoff as you mentioned. And the benefit is we don’t need the efficientPowerOfTwo utility any more.

src/refresh/RefreshTimer.ts
Eskibear: current calculated time is correct. 
Then we probably need to change calculateBackoffTime, making it equal to pow(2, failedAttempts-1) * minBackoff

src/AzureAppConfigurationImpl.ts
Eskibear: Technically we can, because `this.#refreshEnabled` depends on `this.#options` to have the correct result, and it's assigned at the beginning of the constructor. 
However, `this.#options` can be undefined according to its declaration, and TypeScript is not smart enough to know it's not undefined if we use `this.#refreshEnabled`. That would require we to add not-null assertion `!` as below:

```ts
// Currently type deduction works, because in the if block `options.refreshOptions` is not null.
if (options?.refreshOptions?.enabled) {
    const { watchedSettings, refreshIntervalInMs } = options.refreshOptions;

// After the change: no type deduction for this.#options, we have to add assertion, which is a little bit wired.
if (this.#refreshEnabled) {
    const { watchedSettings, refreshIntervalInMs } = this.#options!.refreshOptions!;   // <--  ! mark is not-null assertion
```

Eskibear: Another reason is, `this.#refreshEnabled` is for future calls to know the enablement in elsewhere. Here in the constructor, we initialize related properties according to the passed parameter `options` instead of `this.#options`. It makes sense. 
So I prefer to keep it as is, although it looks like we can re-use it at a glance.

src/AzureAppConfigurationImpl.ts
Eskibear: It's a good question. Per last design and the refresh behavior [above](https://github.com/Azure/AppConfiguration-JavaScriptProvider/pull/21#issue-1962913478), it's "no refresh", meaning we don't support refresh for the case.

I agree with you to explicitly throw an error.

src/AzureAppConfigurationImpl.ts
Eskibear: > If a watched setting is not loaded as part of the kv selectors

Good question. Is that a valid case? To have a dedicated watched setting but not loaded from any selector? I doubt that a little bit from a user's perspective. 

> how do we update its etag in this.sentinels?

Back to the question. For the change-detected sentinel, its etag is updated immediately. But if there are multiple sentinels and since we break the loop and skip sending requests to check rest sentinels, their etags are not updated at that time. They are expected to be updated during the new load here. The consequence in such case is, if I update 2 sentinels at the same time, it might eventually refresh twice, although eventually the etags will be updated.

src/AzureAppConfigurationImpl.ts
Eskibear: Tried and it works as expected. The client returns configuration setting with status code 200.

src/refresh/RefreshTimer.ts
Eskibear: Or we denote the `failedAttempts` with "failed attempts before current one"... Anyway either way is ok for me.

src/AzureAppConfigurationImpl.ts
Eskibear: My initial thought is as below, and we don't mess them up. It's much clearer.
- `settingSelector` specifies what to load
- `watchedSetting` specifies sentinels
Synced with Zhenlan offline, learning that dotnet provider used to be this behavior, but per customer feedback/request, watched settings are also loaded even not specified as part of selectors. 

I'll change to have consistent behavior with dotnet provider, loading watched settings even when they are not part of selectors. In terms of precedence, watched settings will have higher precedence.

src/AzureAppConfigurationImpl.ts
Eskibear: updated in 506c8a6

src/refresh/RefreshTimer.ts
Eskibear: updated in 98b12e2

src/refresh/RefreshTimer.ts
avanigupta: I might be missing something here. Why do we need to calculate failedAttempts-1?
We don't need to calculate back off for the first attempt (2^1) because the first time we can simply back off for the min duration (just an optimisation). 
We do need to calculate back off from the second attempt onwards (2^2).

src/AzureAppConfigurationImpl.ts
Eskibear: implemented in 5d80ccc

examples/README.md
```diff
@@ -39,7 +39,7 @@ To run the examples using the published version of the package:
         ```bash
         npx cross-env APPCONFIG_CONNECTION_STRING="<appconfig connection string>" 
         ```
-
```
avanigupta, line 42: Nit: Changes to this file can be reverted. It only differs in whitespace.

src/AzureAppConfigurationImpl.ts
```diff
@@ -99,19 +274,36 @@ export class AzureAppConfigurationImpl extends Map<string, unknown> implements A
...
     }
+
+    async #getConfigurationSettingWithTrace(configurationSettingId: ConfigurationSettingId, customOptions?: GetConfigurationSettingOptions): Promise<GetConfigurationSettingResponse | undefined> {
+        let response: GetConfigurationSettingResponse | undefined;
+        try {
+            const options = {...customOptions ?? {}};
+            if (this.#requestTracingEnabled) {
+                const requestType = this.#isInitialLoadCompleted ? RequestType.Watch : RequestType.Startup;
+                options.requestOptions = {
+                    customHeaders: this.#customHeaders(requestType)
```
avanigupta, line 294: Since we added this new variable `isInitialLoadCompleted`, we can simplify wherever `requestType` is passed. We only need to check `isInitialLoadCompleted` while calling `createCorrelationContextHeader` now.

avanigupta: Or maybe we can get rid of `#customHeaders` and directly call `createCorrelationContextHeader` wherever we want to get custom headers.

avanigupta:
I left some comments on this commit but they're not showing up on the main PR for some reason: https://github.com/Azure/AppConfiguration-JavaScriptProvider/commit/5d80ccca527000e6cb2083f93557172d28463582

src/AzureAppConfigurationImpl.ts
Eskibear: good idea. I can start with changing `#customHeaders(requestType: RequestType)` to `#customHeaders()`, and check isInitialLoadCompleted in its implmentation. 
`createCorrelationContextHeader(...)` is a util function assembling correlation context header, which I would prefer to keep it as is, accepting request type as an input.

src/AzureAppConfigurationImpl.ts
```diff
@@ -59,26 +111,149 @@ export class AzureAppConfigurationImpl extends Map<string, unknown> implements A
...
+    }
+
+    async #loadSelectedAndWatchedKeyValues() {
+        const keyValues: [key: string, value: unknown][] = [];
+
+        const loadedSettings = await this.#loadSelectedKeyValues();
+        const watchedSettings = await this.#loadWatchedKeyValues(loadedSettings);
+
+        // process key-values, watched settings have higher priority
+        for (const setting of [...loadedSettings.values(), ...watchedSettings.values()]) {
```
Eskibear, line 167: > We need to load sentinel keys just to watch their etags, but we dont need to add those sentinel keys to the configuration.   --Avani

src/AzureAppConfigurationImpl.ts
Eskibear: I'll also change `createCorrelationContextHeader(options, requestType)` to `createCorrelationContextHeader(options, isInitialLoadCompleted)` as suggested, which makes sense.

examples/README.md
Eskibear: reverted by d838076

src/AzureAppConfigurationImpl.ts
```diff
@@ -1,28 +1,46 @@
...
+import { KeyFilter, LabelFilter, SettingSelector } from "./types";
 
-export class AzureAppConfigurationImpl extends Map<string, unknown> implements AzureAppConfiguration {
+/**
+ * Key-value identifier, used as key of a Map.
+ * A primitive string type is actually used, because operator === is used to compare identifiers and we cannot override it.
+ */
+type KeyValueIdentifier = string; // key::label
+function toKeyValueIdentifier(key: string, label: string | undefined): KeyValueIdentifier {
+    return `${key}::${label ?? ""}`;
```
Eskibear, line 24: > This approach might produce same KeyValueIdentifier for different key-label combinations. For eg, these will have the same output:
> 
> key="AppName" and label="Dev"
> key="AppName::Dev" and label=""
> 
> Can we use a different model to uniquely identify key/label combinations?
> -- Avani

Eskibear: @avanigupta 
> key="AppName" and label="Dev", id is "AppName::Dev"
> key="AppName::Dev" and label="", id is "AppName::Dev::"

Unproper example, they are not the same. 

Another example, in below cases both have id "A::B::C".
- key="A::B" label="C"
- key="A"  label="B::C"

Eskibear: Using such simple formatted string will always have corner cases uncovered.

Trying making it more complicated, e.g. using `key:${key},label:${label}` as id, we might cover most **valid** cases. But we can still construct corner case as below:
`key`="A,label:B", `label`="C", `id`="key:A,label:B,label:C"
`key`="A", `label`="B,label:C", `id`="key:A,label:B,label:C"

Eskibear: Googled and found some are using `JSON.stringify()`.  But it doesn't solve the problem.
1. `{key: a, label:b}` and `{label:b, key:a}` are serialized into different strings, which is not expected. But this can be solved by sorting properties by keys, and we don't have nested values here.
2. `{key: a, label:b}` is serialized as a string `{"key":"a","label":"b"}`, but essentially it's also a type of formatted string, with `{"key":"` as prefix, `","label":"` as separator, `"}` as postfix. Counter-example can be easily constructed as above.

avanigupta: Why do we have to use strings at all? Is it possible to create a new type with key and label fields, and a custom comparator to define equality? I'm imagining something like our .NET [KeyValueIdentifier](https://github.com/Azure/AppConfiguration-DotnetProvider/blob/main/src/Microsoft.Extensions.Configuration.AzureAppConfiguration/Models/KeyValueIdentifier.cs)

Eskibear: There is currently a [proposal](https://github.com/tc39/proposal-record-tuple#usage-in-mapsetweakmapweakset) to add Records and Tuples (basically immutable Objects and Arrays) to JavaScript.

So I would suggest:
a) either we use a relatively complicated separator for formatted string, e.g. as above, `key:${key},label:${label}`.
b) or we don't use the Map, but an Array, and iterate the array to find the matched element every time (of course more cost)...


WDYT? @avanigupta

src/AzureAppConfigurationImpl.ts
Eskibear: OK, let's not include watched settings as discussed in the thread.

src/AzureAppConfigurationImpl.ts
Eskibear: > and a custom comparator to define equality?

No, it's impossible in JavaScript...(yes, today, in 2023). It forces to use `===` to compare keys, and we cannot override operator in JavaScript.
FYI. https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Map#key_equality

src/AzureAppConfigurationImpl.ts
Eskibear: updated by 97e0350

src/AzureAppConfigurationImpl.ts
avanigupta: > a) either we use a relatively complicated separator for formatted string, e.g. as above, `key:${key},label:${label}`. 

I am hesitant to use this approach because it will always leave room for some error.

> b) or we don't use the Map, but an Array, and iterate the array to find the matched element every time (of course more cost)...

This doesn't sound too bad. From the current code, it looks like we're not really searching in the map anyway. We're mostly using the map iteratively. Moreover, since the map only holds watched settings, and we know that users only provide a few sentinel keys (it's difficult to setup 100s of watched settings) - iterating the array should be fine too.

Eskibear: Agree. Let’s use array. We should ensure the correctness first, and improve the performance later if necessary.

Eskibear: fixed in d0440dc, changing sentinels to array

src/refresh/RefreshTimer.ts
```diff
@@ -0,0 +1,82 @@
...
+ *  - If the interval is between the minimum and maximum backoff, the interval is used as the maximum backoff.
+ *  - Because of the jitter, the maximum backoff time is actually `MaximumBackoffInMs` * (1 + `JitterRatio`).
+ */
+
+const MinimumBackoffInMs = 30 * 1000; // 30s
+const MaximumBackoffInMs = 10 * 60 * 1000; // 10min
+const MaxSafeExponential = 30; // Used to avoid overflow. bitwise operations in JavaScript are limited to 32 bits. It overflows at 2^31 - 1.
+const JitterRatio = 0.25;
+
+export class RefreshTimer {
```
avanigupta, line 24: Is there a way to make this class internal such that our provider is the only consumer of its methods?

src/AzureAppConfigurationImpl.ts
```diff
@@ -60,25 +105,141 @@ export class AzureAppConfigurationImpl extends Map<string, unknown> implements A
...
+        }
+
+        // try refresh if any of watched settings is changed.
+        let needRefresh = false;
+        for (const sentinel of this.#sentinels.values()) {
+            const response = await this.#getConfigurationSettingWithTrace(sentinel, {
+                onlyIfChanged: true
+            });
+
+            if (response === undefined || response.statusCode === 200) {
```
avanigupta, line 196: What if the sentinel was initially not loaded, and its still not present on the server?
response will be undefined, but we should not trigger refresh because nothing has changed.

src/refresh/RefreshTimer.ts
Eskibear: Currently it is kind of internal, because we don't export it in `src/index.ts`.  Here `export` means you can reference it in other files/modules.
For users using CJS module system, it's internal, no way to import "RefreshTimer" in user code. 
For users using ESM module system, they can import RefreshTimer from "path-to-refreshTimer.ts" if they insists.

If you want to ensure it's 100% internal, we can move it into the file using it, and remove `export`. But that's not a big issue. I would rather keep it as a single file for readability...

avanigupta: Sounds good, lets keep it separate

src/AzureAppConfigurationImpl.ts
Eskibear: that's exactly a case not covered, as in the comment says in the next line.  
- changed [covered]
- deleted [covered]
- created [not covered]

We should cover it. And my idea is:
Compare sentinel.etag with response?.etag, if not equal, then need refresh.

Cases should all be covered:
- changed: etag are different.
- deleted: sentinel.etag has value but response is undefined
- created: sentinel.etag is undefined, but response.etag has value

WDYT?

avanigupta: Actually I think created case is covered. The case not being covered is when a sentinel key is "Not Created".
We only need to check the etag when response is undefined. 

changed: etag are different, so response will be 200 (already covered)
deleted: sentinel.etag has value but response is undefined (already covered)
created: sentinel.etag is undefined, but response.etag has value, so response will be 200 (already covered)
NOT created: sentinel.etag is undefined, and response is also undefined -> Not covered

So you can just check if sentinel.etag is undefined when response is undefined -> this condition should not trigger a refresh.

Eskibear: fixed in 8acf87b, according to the idea above.

Eskibear: yes, as you mentioned, "created" WAS already covered. The commit fixes the case "not created".

Eskibear: > just check if sentinel.etag is undefined when response is undefined

It's actually identical to checking `sentinel.etag === response?.etag`. Although they should never equal if they both have values (otherwise 304 and not go into this block).

src/AzureAppConfigurationImpl.ts
```diff
@@ -60,25 +105,143 @@ export class AzureAppConfigurationImpl extends Map<string, unknown> implements A
...
+                // sentinel deleted / changed / created.
+                if (sentinel.etag !== response?.etag) {
```
avanigupta, line 198: if statements can be combined:

`if ((response === undefined && sentinel.etag !=== undefined) || response.statusCode === 200) {`

avanigupta (approved):

src/AzureAppConfigurationImpl.ts
Eskibear: simplified in fd60862

avanigupta (approved):

zhiyuanliang-ms (approved):

PR #22, opened 2023-11-02
forbid wildcard in label filters

Eskibear:
This PR:
- forbid wildcards by throwing errors if `*` or `,` detected in label filters. (because we cannot handle settings with the same key but different labels"
- update test utils and add corresponding test cases.

### DIFF
--- a/src/AzureAppConfigurationImpl.ts
+++ b/src/AzureAppConfigurationImpl.ts
@@ -46,7 +46,7 @@ export class AzureAppConfigurationImpl extends Map<string, unknown> implements A
         const keyValues: [key: string, value: unknown][] = [];
 
         // validate selectors
-        const selectors = validatedSelectors(this.options?.selectors);
+        const selectors = getValidSelectors(this.options?.selectors);
 
         for (const selector of selectors) {
             const listOptions: ListConfigurationSettingsOptions = {
@@ -109,7 +109,7 @@ export class AzureAppConfigurationImpl extends Map<string, unknown> implements A
     }
 }
 
-function validatedSelectors(selectors?: { keyFilter: string, labelFilter?: string }[]) {
+function getValidSelectors(selectors?: { keyFilter: string, labelFilter?: string }[]) {
     if (!selectors || selectors.length === 0) {
         // Default selector: key: *, label: \0
         return [{ keyFilter: KeyFilter.Any, labelFilter: LabelFilter.Null }];
@@ -117,7 +117,7 @@ function validatedSelectors(selectors?: { keyFilter: string, labelFilter?: strin
     return selectors.map(selectorCandidate => {
         const selector = { ...selectorCandidate };
         if (!selector.keyFilter) {
-            throw new Error("Key filters cannot be empty.");
+            throw new Error("Key filter cannot be null or empty.");
         }
         if (!selector.labelFilter) {
             selector.labelFilter = LabelFilter.Null;
--- a/src/AzureAppConfigurationImpl.ts
+++ b/src/AzureAppConfigurationImpl.ts
@@ -44,7 +44,10 @@ export class AzureAppConfigurationImpl extends Map<string, unknown> implements A
 
     public async load() {
         const keyValues: [key: string, value: unknown][] = [];
-        const selectors = this.options?.selectors ?? [{ keyFilter: KeyFilter.Any, labelFilter: LabelFilter.Null }];
+
+        // validate selectors
+        const selectors = validatedSelectors(this.options?.selectors);
+
         for (const selector of selectors) {
             const listOptions: ListConfigurationSettingsOptions = {
                 keyFilter: selector.keyFilter,
@@ -104,4 +107,24 @@ export class AzureAppConfigurationImpl extends Map<string, unknown> implements A
         headers[CorrelationContextHeaderName] = this.correlationContextHeader;
         return headers;
     }
+}
+
+function validatedSelectors(selectors?: { keyFilter: string, labelFilter?: string }[]) {
+    if (!selectors || selectors.length === 0) {
+        // Default selector: key: *, label: \0
+        return [{ keyFilter: KeyFilter.Any, labelFilter: LabelFilter.Null }];
+    }
+    return selectors.map(selectorCandidate => {
+        const selector = { ...selectorCandidate };
+        if (!selector.keyFilter) {
+            throw new Error("Key filters cannot be empty.");
+        }
+        if (!selector.labelFilter) {
+            selector.labelFilter = LabelFilter.Null;
+        }
+        if (selector.labelFilter.includes("*") || selector.labelFilter.includes(",")) {
+            throw new Error("The characters '*' and ',' are not supported in label filters.");
+        }
+        return selector;
+    });
 }
--- a/src/AzureAppConfigurationOptions.ts
+++ b/src/AzureAppConfigurationOptions.ts
@@ -9,10 +9,23 @@ export const MaxRetryDelayInMs = 60000;
 
 export interface AzureAppConfigurationOptions {
     /**
-     * Selectors used to control what key-values are retrieved from Azure App Configuration.
+     * Specify what key-values to include in the configuration provider.  include multiple sets of key-values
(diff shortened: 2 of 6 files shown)


src/AzureAppConfigurationImpl.ts
```diff
@@ -105,3 +108,23 @@ export class AzureAppConfigurationImpl extends Map<string, unknown> implements A
...
 }
+
+function validatedSelectors(selectors?: { keyFilter: string, labelFilter?: string }[]) {
+    if (!selectors || selectors.length === 0) {
+        // Default selector: key: *, label: \0
+        return [{ keyFilter: KeyFilter.Any, labelFilter: LabelFilter.Null }];
+    }
+    return selectors.map(selectorCandidate => {
+        const selector = { ...selectorCandidate };
+        if (!selector.keyFilter) {
```
avanigupta, line 119: Does this check make sure that key is not null or empty?

src/AzureAppConfigurationImpl.ts
```diff
@@ -105,3 +108,23 @@ export class AzureAppConfigurationImpl extends Map<string, unknown> implements A
...
+            throw new Error("Key filters cannot be empty.");
```
avanigupta, line 120:
```suggestion
            throw new Error("Key filter cannot be null or empty.");

```

src/AzureAppConfigurationImpl.ts
```diff
@@ -105,3 +108,23 @@ export class AzureAppConfigurationImpl extends Map<string, unknown> implements A
         return headers;
     }
 }
+
+function validatedSelectors(selectors?: { keyFilter: string, labelFilter?: string }[]) {
+    if (!selectors || selectors.length === 0) {
+        // Default selector: key: *, label: \0
+        return [{ keyFilter: KeyFilter.Any, labelFilter: LabelFilter.Null }];
+    }
+    return selectors.map(selectorCandidate => {
```
avanigupta, line 117: Will this return unique selectors if there are any duplicates?

src/AzureAppConfigurationImpl.ts
```diff
@@ -44,7 +44,10 @@ export class AzureAppConfigurationImpl extends Map<string, unknown> implements A
 
     public async load() {
         const keyValues: [key: string, value: unknown][] = [];
-        const selectors = this.options?.selectors ?? [{ keyFilter: KeyFilter.Any, labelFilter: LabelFilter.Null }];
+
+        // validate selectors
+        const selectors = validatedSelectors(this.options?.selectors);
```
avanigupta, line 49: `validatedSelectors` sounds more like a variable name than a method name.
What do you think about calling it `getValidSelectors`?

src/AzureAppConfigurationOptions.ts
```diff
@@ -8,7 +8,13 @@ export const MaxRetries = 2;
...
 
 export interface AzureAppConfigurationOptions {
-    selectors?: { keyFilter: string, labelFilter: string }[];
+    /**
+     * Selectors used to control what key-values are retrieved from Azure App Configuration.
+     *
+     * @property keyFilter: A filter that determines the set of keys that are included in the configuration provider, cannot be omitted.
+     * @property labelFilter: A filter that determines what label to use when selecting key-values for the the configuration provider, omitted for no label.
+     */
+    selectors?: { keyFilter: string, labelFilter?: string }[];
```
avanigupta, line 17: Can you use the same description that we have in [.NET provider](https://github.com/Azure/AppConfiguration-DotnetProvider/blob/75f3d6c186bbfb8c43e62f37207b1193e66e39e6/src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationOptions.cs#L117-L136)? That summary has been iterated over a few times based on customer feedback and the same rules apply here too.

src/AzureAppConfigurationImpl.ts
Eskibear: Yes. `keyFilter` is of string type, so the condition here is to ensure `keyFilter` is not `undefined` or `null` or `""` (empty string).

BTW for Javascript language, you actually can pass anything in if you insist, and here it also ensures it's not number `0` etc. But that's already guarded by the type check at compile time for TypeScript, and not the case we need to take care.
See https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Boolean#boolean_coercion

More info in case you are interested in. See https://stackoverflow.com/a/154068 for various ways to detect empty strings. To me, the most accurate way to translate "is null or empty string" might be as below, but seldomly used because it's complicated. And we have typescript now to guard unexpected inputs.
```js
function isNullOrEmptyString(str) {
  return str === undefined
     || str === null
     || typeof str === "string" &&  str.length === 0
}
```

src/AzureAppConfigurationImpl.ts
Eskibear: In fact the type definition `keyFilter: string` guards that it cannot be `null` (if strictly checked). But again it won't work if type check is disabled/skipped. 

So yes, I will commit you suggestion to make it more clear.

src/AzureAppConfigurationImpl.ts
Eskibear: No, but the settings are ensured to be unique, as they are overridden by the last selector if there are any duplicates or overlaps among selectors.

We cannot identify overlaped selectors and combine them into one, but yes we can dedup the selectors with exactly the same keyFilter and labelFilter, to save requests (if they do specify the same selectors multiple times).

src/AzureAppConfigurationOptions.ts
Eskibear: sure, and maybe later I will probably create a type alias for `{ keyFilter: string, labelFilter?: string }`, e.g. `KeyValueSelector`.

src/AzureAppConfigurationImpl.ts
avanigupta: > but yes we can dedup the selectors with exactly the same keyFilter and labelFilter, to save requests (if they do specify the same selectors multiple times).

Yes, that's the case I was talking about. Dedup exactly same key and label filters to save requests.

Eskibear: @avanigupta I have a question. Since the order of selectors also reflects priority, what would be the effective selector for this case?
I think current impl in .NET provider has the same problem.
```ts
selectors: [ 
            { keyFilter: "app.*", labelFilter: "dev" },
            { keyFilter: "app.*", labelFilter: "production" },
            { keyFilter: "app.*", labelFilter: "dev" },
] 
```

Without dedup, it's clear that the last "app.* + dev" wins. But with simple dedup, the last selector is not counted in, the effective one becomes  "app.* + production". I think current impl in .NET provider has the same problem.

Do you think we should reverse the array before dedup?

avanigupta: That's a good catch. This looks like a [bug in the .NET provider](https://github.com/Azure/AppConfiguration-DotnetProvider/blob/75f3d6c186bbfb8c43e62f37207b1193e66e39e6/src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationOptions.cs#L154C27-L154C27). We need to deduplicate while maintaining the priority of the last occurrence. In this case, "app.* + dev" should always win.

cc @jimmyca15 @amerjusupovic

Eskibear: Shall I create a separate issue in .NET provider repo?

avanigupta: > Shall I also create a separate issue in .NET provider repo?

Sounds like a good idea.

Eskibear: @avanigupta 
Here is the issue for .NET provider: https://github.com/Azure/AppConfiguration-DotnetProvider/issues/490

BTW, I also created an issue here in JS provider, and we can add the dedup logic correctly later in a separate PR dedicated for that purpose. For this PR, I would like to get it merged as it is since the behavior is as expected. WDYT?

avanigupta (approved):

PR #31, opened 2023-11-27
Dedup exact same selectors while keep original precedence

### LINKED ISSUE #23: Dedup exact same selectors while keep original precedence
Dedup same selectors can save requests. And we should ensure the precedence of originally passed selectors unchanged.

```ts
selectors: [ 
            { keyFilter: "app.*", labelFilter: "dev" },
            { keyFilter: "app.*", labelFilter: "production" },
            { keyFilter: "app.*", labelFilter: "dev" },
] 
```

Without dedup, it's clear that the last "app.* + dev" wins. After dedup, we should keep the same precedence, and in this case let "app.* + dev" win.

_Originally posted by @Eskibear in https://github.com/Azure/AppConfiguration-JavaScriptProvider/pull/22#discussion_r1380932655_

Eskibear:
fix #23 

The impl is to reverse the selectors, dedup, and reverse it again.

With a little refactoring, extracting LabelFilter/KeyFilter/SettingSelector into types.ts.


src/AzureAppConfigurationImpl.ts
```diff
@@ -109,12 +109,23 @@ export class AzureAppConfigurationImpl extends Map<string, unknown> implements A
...
+function getValidSelectors(selectors?: SettingSelector[]) {
     if (!selectors || selectors.length === 0) {
         // Default selector: key: *, label: \0
         return [{ keyFilter: KeyFilter.Any, labelFilter: LabelFilter.Null }];
     }
-    return selectors.map(selectorCandidate => {
+
+    // below code dedupes selectors by keyFilter and labelFilter, the latter selector wins
+    const dedupedSelectors: SettingSelector[] = [];
+    const reversedSelectors = [...selectors].reverse();
```
avanigupta, line 120: Is there a reason for choosing list reversal logic here? Since we have the same code in .NET, how about keeping the logic consistent?

Eskibear: No specific reason, just for convenience, as reverse() is in-place. I'll use keep the logic consistent with .NET.

avanigupta (approved):

PR #35, opened 2023-12-14
Add error handling delay in load function

Eskibear:
load() method is called in the application's startup code path.
Unhandled exceptions cause application crash which can result in crash loops as orchestrators attempt to restart the application.
Knowing the intended usage of the provider in startup code path, we mitigate back-to-back crash loops from overloading the server with requests by waiting a minimum time to propagate fatal errors.

### DIFF
--- a/src/load.ts
+++ b/src/load.ts
@@ -34,7 +34,7 @@ export async function load(
     } catch (error) {
         // load() method is called in the application's startup code path.
         // Unhandled exceptions cause application crash which can result in crash loops as orchestrators attempt to restart the application.
-        // Knowing the intended usage of the provider in startup code path, we mitigate back-to-back crash loops from overloading the server with requests by waiting a minimum time to propogate fatal errors.
+        // Knowing the intended usage of the provider in startup code path, we mitigate back-to-back crash loops from overloading the server with requests by waiting a minimum time to propagate fatal errors.
         const delay = MinDelayForUnhandedError - (Date.now() - startTimestamp);
         if (delay > 0) {
             await new Promise((resolve) => setTimeout(resolve, delay));
--- a/src/load.ts
+++ b/src/load.ts
@@ -29,27 +29,10 @@ export async function load(
     appConfigOptions?: AzureAppConfigurationOptions
 ): Promise<AzureAppConfiguration> {
     const startTimestamp = Date.now();
-    try {
-        return await loadPromise(connectionStringOrEndpoint, credentialOrOptions, appConfigOptions);
-    } catch (error) {
-        // load() method is called in the application's startup code path.
-        // Unhandled exceptions cause application crash which can result in crash loops as orchestrators attempt to restart the application.
-        // Knowing the intended usage of the provider in startup code path, we mitigate back-to-back crash loops from overloading the server with requests by waiting a minimum time to propagate fatal errors.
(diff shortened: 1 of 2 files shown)


juniwang (approved):

src/load.ts
```diff
@@ -25,6 +27,26 @@ export async function load(
     connectionStringOrEndpoint: string | URL,
     credentialOrOptions?: TokenCredential | AzureAppConfigurationOptions,
     appConfigOptions?: AzureAppConfigurationOptions
+): Promise<AzureAppConfiguration> {
+    const startTimestamp = Date.now();
+    try {
+        return await loadPromise(connectionStringOrEndpoint, credentialOrOptions, appConfigOptions);
+    } catch (error) {
```
avanigupta, line 34: I think we want to avoid crash loops for server-side errors like 401/403. But client-side errors like input validation (eg: invalid connection string) can be thrown instantly. Is there a way to filter out known client-side errors and fail fast?

cc @mrm9084 since I see the same issue in your PR for Python provider.

Eskibear: I just revisited dotnet's impl and find it throws ArgumentException immediately. 
```c#
catch (ArgumentException) 
{
    // Instantly re-throw the exception
    throw;
}
```

For JavaScript I think of two approaches for the moment, and I prefer option b). 
In load.ts it does the input validation and in AzureAppConfigurationImpl.ts it sends requests to server.

option a) move the delay logic into `AzureAppConfigurationImpl.load()`, but since it's also called by dynamic refresh, so later we probably also check if it's from a refresh...

option b) in load.ts we don't wrap[ the validation part](https://github.com/Azure/AppConfiguration-JavaScriptProvider/blob/e8292076d5d25f9b8aa17f4200434c905c19958c/src/load.ts#L31-L56) with the delay logic, and only wrap [the part sending requests](https://github.com/Azure/AppConfiguration-JavaScriptProvider/blob/e8292076d5d25f9b8aa17f4200434c905c19958c/src/load.ts#L58-L60). 
BTW technically we can create a new type ArgumentError and throw errors of this type in the validation part. But since the validation part is in a single place, it's simpler not to invent a new error type but just exclude the part.

avanigupta: Option B sounds good then!

avanigupta (approved):

src/load.ts
mrm9084: I'm not sure if be 100% solves the issue. I'm not sure about in the java script sdk. But the Key Vault clients can't be made until keys are load if they only provider a credential. In this code the wait will trigger on a bad key vault credential.

avanigupta: >  In this code the wait will trigger on a bad key vault credential.

I think that's intentional. If the KeyVault credential is wrong, we want the wait to trigger - simply because if we don't wait, provider will continue to make AppConfig calls and fail on a loop. We want to avoid any calls to AppConfig in a tight loop.

mrm9084: A failed call to app config and a successful call to app config are the same number of requests.

mrm9084: This would only matter it they are loading over 100 keys I guess.

mrm9084: Discussed offline, we should be good.

mrm9084 (approved):

src/load.ts
avanigupta: Discussed offline with Matt. This PR should be good to merge.

Issue #29, opened 2023-11-21
Adopt private field/methods in ES2022


Eskibear:
https://262.ecma-international.org/13.0/#sec-privateelement-specification-type
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Classes/Private_properties

We can prepend `#` to names of private fields and methods. Then users won't have chance to access those internal members.

PR #37, opened 2024-01-02
hide private properties of AppConfigurationImpl

Eskibear:
See #29 

Previously, TS keyword `private` only during compilation stage. In JS you can still access "private members" if you insist. 
This PR adopts private properties from ECMAScript, forbidding the access of private members, for JavaScript in runtime.

Code changes:
- private members => prepend `#` and remove `private`
- public members => remove `public` which is the default scope.

### DIFF
--- a/src/AzureAppConfigurationImpl.ts
+++ b/src/AzureAppConfigurationImpl.ts
@@ -13,58 +13,63 @@ import { createCorrelationContextHeader, requestTracingEnabled } from "./request
 import { SettingSelector } from "./types";
 
 export class AzureAppConfigurationImpl extends Map<string, unknown> implements AzureAppConfiguration {
-    private adapters: IKeyValueAdapter[] = [];
+    #adapters: IKeyValueAdapter[] = [];
     /**
      * Trim key prefixes sorted in descending order.
      * Since multiple prefixes could start with the same characters, we need to trim the longest prefix first.
      */
-    private sortedTrimKeyPrefixes: string[] | undefined;
-    private readonly requestTracingEnabled: boolean;
-    private correlationContextHeader: string | undefined;
+    #sortedTrimKeyPrefixes: string[] | undefined;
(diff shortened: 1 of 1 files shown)


Eskibear:
E.g. try accessing private member `client`.
```js
const settings = await load(connectionString);
```

Before:
image


After:
image

zhiyuanliang-ms:
Will this change raise any concerns about compatibility?

src/AzureAppConfigurationImpl.ts
```diff
@@ -13,58 +13,63 @@ import { createCorrelationContextHeader, requestTracingEnabled } from "./request
...
+        this.#adapters.push(new JsonKeyValueAdapter());
     }
 
-    public async load() {
+    async load() {
         const keyValues: [key: string, value: unknown][] = [];
 
         // validate selectors
-        const selectors = getValidSelectors(this.options?.selectors);
+        const selectors = getValidSelectors(this.#options?.selectors);
```
zhiyuanliang-ms, line 54: I think we'd better make getValidSelector a private static method or it should not be included in this file.

zhiyuanliang-ms (approved):

src/AzureAppConfigurationImpl.ts
Eskibear: I made it top-level method for simplicity. No much difference in JavaScript. But I'm ok to make is a private static method here.

Eskibear: No, I decide to leave it as top-level functions at least for now. It would be tedious when I call static private functions as below:
```js
// within class AzureAppConfigurationImpl
const selectors = AzureAppConfigurationImpl.getValidSelectors(this.options?.selectors);
```

zhiyuanliang-ms:
Should we unify the public/private method in Adapters?

Eskibear:
> Should we unify the public/private method in Adapters?

Yes. Tracked in #29

PR #43, opened 2024-01-03
Update labelFilter default value in SettingSelector

Eskibear:
image


zhiyuanliang-ms:
Reference: https://tsdoc.org/pages/tags/defaultvalue/

zhiyuanliang-ms (approved):

PR #44, opened 2024-01-04
Rename AzureAppConfigurationKeyVaultOptions to KeyVaultOptions

Eskibear:


zhiyuanliang-ms (approved):

src/AzureAppConfigurationOptions.ts
```diff
@@ -2,7 +2,7 @@
 // Licensed under the MIT license.
 
 import { AppConfigurationClientOptions } from "@azure/app-configuration";
-import { AzureAppConfigurationKeyVaultOptions } from "./keyvault/AzureAppConfigurationKeyVaultOptions";
+import { KeyVaultOptions } from "./keyvault/AzureAppConfigurationKeyVaultOptions";
```
avanigupta, line 5: Should we change the name of the file?

Eskibear: Yes. Good catch. ;)

avanigupta (approved):

PR #45, opened 2024-01-30
Add all-in-one pack script

Eskibear:
This script helps to simplify the definition of build/release pipelines. It builds and creates tarball pack in the project root, no matter where you execute the script.

### DIFF
--- a/scripts/build-and-pack.sh
+++ b/scripts/build-and-pack.sh
@@ -1,0 +1,24 @@
+#!/bin/bash
+
+# Stop on error.
+set -e
+
+# Echo commands.
+set -x
+
+# Get the directory of the script.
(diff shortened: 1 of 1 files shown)


scripts/build-and-pack.sh
```diff
@@ -0,0 +1,24 @@
+#!/bin/bash
+
+# Stop on error.
+set -e
+
+# Echo commands.
+set -x
```
juniwang, line 7: `set -x` is for debug purpose in my mind. I would suggest removing it in official code to prevent printing unexpected content in future.

Eskibear: good point. I'll explicit echo commands instead.

juniwang (approved):

PR #46, opened 2024-01-31
rename members to adopt private properties

Eskibear:
See #29 
Use ECMAScript private for better guarantee of the property isolation.

private => remove "private" and prepend `#` .
public => remove "public" as it's the default.


src/common/disposable.ts
```diff
@@ -2,14 +2,17 @@
 // Licensed under the MIT license.
 
 export class Disposable {
-    private disposed = false;
-    constructor(private callOnDispose: () => any) { }
+    #disposed = false;
+    #callOnDispose: () => any;
+
+    constructor(callOnDispose: () => any) {
```
juniwang, line 8: Can client pass in null or undefined to the constructor? it's all good if no.

juniwang:
Is this private property standard widely adopted by JS libraries? Do we know the usage of old version browsers that do not support it?

Eskibear:
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Classes/Private_properties#browser_compatibility
It's supported by almost all popular browser/runtimes in 2021. So I don't worry about compatibility.

Eskibear:
As for how widely adopted, as far as I can tell there is quite a portion from[ a yearly survey](https://2022.stateofjs.com/en-US/features/language/#private_fields): 
image

And some repos under Azure org are starting to adopt it after a quick search, e.g. https://github.com/Azure/cadl-ranch/blob/c611311c3d36a3048a3077137cbd4e6be762d034/packages/cadl-ranch-coverage-sdk/src/client.ts#L12

src/common/disposable.ts
zhiyuanliang-ms: Technically, it is possible. Using "any" or optional type can pass the static type check. It would be better to validate `callOnDispose`'s type here.

zhiyuanliang-ms (approved):

src/common/disposable.ts
Eskibear: The answer is no. @juniwang 
()=>any is a function whose return type is any. Undefined or null cannot be converted to a function type. @zhiyuanliang-ms

PR #48, opened 2024-02-06
Assert error messages and increase timeout for tests

Eskibear:
Ref:
[chai-promise](https://www.chaijs.com/plugins/chai-as-promised/) supports to also assert error message when a promise is rejected.

### DIFF
--- a/test/load.test.ts
+++ b/test/load.test.ts
@@ -31,6 +31,8 @@ const mockedKVs = [{
 }].map(createMockedKeyValue);
 
 describe("load", function () {
+    this.timeout(10000);
+
     before(() => {
         mockAppConfigurationClientListConfigurationSettings(mockedKVs);
     });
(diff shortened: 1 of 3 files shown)


test/load.test.ts
```diff
@@ -117,23 +119,26 @@ describe("load", function () {
         expect(settings.get("KeyForEmptyValue")).eq("");
     });
 
-    it("should not support * or , in label filters", async () => {
+    it("should not support * in label filters", async () => {
```
zhiyuanliang-ms, line 122: ',' is mentioned below, is it removed deliberately?
```
return expect(loadWithWildcardLabelFilter).to.eventually.rejectedWith("The characters '*' and ',' are not supported in label filters.");
```

zhiyuanliang-ms (approved):

PR #49, opened 2024-02-06
Refactor AzureAppConfiguration with new interfaces

Eskibear:
#### Design
We support both two approaches to access key-values, maximize compatibility with customer's application code. For easier adoption of Azure App Configuration, and less code change.

```ts
AzureAppConfiguration {
  get<T>(key: string): T | undefined;  // preferred method, e.g. config.get("app.settings.color")
    constructConfigurationObject(options);    // helper function to construct hierarchical object, compatible with chained dot notation, e.g. config.app.settings.color
}
```

#### Edge case (for using configuration object only):
```
Hierarchical key-value pairs with overlapped key prefix:
  key: "app3.settings" => value: "placeholder"
  key: "app3.settings.fontColor" => value: "yellow"

How to re-construct data object from key-values?
config.app3.settings = ? 
```
In this case, constructConfigurationObject() throws an error.

#### Mitigation approach
~~Add an option `skipObjectValidataion: boolean`, default to `false`.~~
~~- `true`. Throw an error when constructing the `data` object., because of ambiguity.~~
~~- `false`. Skip the validation, allow information lost in `data` object, as the preferred `get()` method can always have the expected results.~~
(propose to use `constructConfigurationObject(options)` as below, which would not affect the original `load()` call)

### DIFF
--- a/src/AzureAppConfiguration.ts
+++ b/src/AzureAppConfiguration.ts
@@ -32,18 +32,6 @@ export interface ConfigurationObjectConstructionOptions {
      * Default is '.'.
      */
     separator?: string;
-
-    /**
-     * The prefix of hierarchical keys to be converted object properties, usefull when converting a subset of the keys.
-     * Default is '', representing the root of the object.
-     */
-    prefix?: string;
-
-    /**
-     * The behavior when error or amibiguity occurs on converting hierarchical keys.
-     * Default is 'error'.
-     */
-    onError?: "error" | "ignore";
 }
 
 interface IGettable {
--- a/src/AzureAppConfiguration.ts
+++ b/src/AzureAppConfiguration.ts
@@ -16,17 +16,17 @@ export type AzureAppConfiguration = {
      * @param thisArg - Optional. Value to use as `this` when executing callback.
      */
     onRefresh(listener: () => any, thisArg?: any): Disposable;
-} & IGettable & IHierarchicalData;
+} & IGettable & IConfigurationObject;
 
-interface IHierarchicalData {
+interface IConfigurationObject {
     /**
-     * Convert the Map-styled data structure to hierarchical object properties.
+     * Construct configuration object based on Map-styled data structure and hierarchical keys.
      * @param options - The options to control the conversion behavior.
      */
-    toHierarchicalData(options?: HierarchicalDataConversionOptions): Record<string, any>;
+    constructConfigurationObject(options?: ConfigurationObjectConstructionOptions): Record<string, any>;
 }
 
-export interface HierarchicalDataConversionOptions {
+export interface ConfigurationObjectConstructionOptions {
     /**
      * The separator to use when converting hierarchical keys to object properties.
      * Default is '.'.
--- a/src/AzureAppConfiguration.ts
+++ b/src/AzureAppConfiguration.ts
@@ -29,10 +29,10 @@ interface IConfigurationObject {
 export interface ConfigurationObjectConstructionOptions {
     /**
      * The separator to use when converting hierarchical keys to object properties.
+     * Supported values: '.', ',', ';', '-', '_', '__', '/', ':'.
      * If separator is undefined, '.' will be used by default.
-
      */
-    separator?: string;
+    separator?: "." | "," | ";" | "-" | "_" | "__" | "/" | ":";
 }
 
 interface IGettable {
--- a/src/AzureAppConfiguration.ts
+++ b/src/AzureAppConfiguration.ts
@@ -16,12 +16,40 @@ export type AzureAppConfiguration = {
      * @param thisArg - Optional. Value to use as `this` when executing callback.
      */
     onRefresh(listener: () => any, thisArg?: any): Disposable;
-} & IGettable & AppConfigurationData;
+} & IGettable & IHierarchicalData;
 
-interface AppConfigurationData {
-    data: { [key: string]: any };
+interface IHierarchicalData {
+    /**
+     * Convert the Map-styled data structure to hierarchical object properties.
+     * @param options - The options to control the conversion behavior.
+     */
+    toHierarchicalData(options?: HierarchicalDataConversionOptions): Record<string, any>;
+}
+
+export interface HierarchicalDataConversionOptions {
+    /**
+     * The separator to use when converting hierarchical keys to object properties.
+     * Default is '.'.
+     */
+    separator?: string;
+
+    /**
+     * The prefix of hierarchical keys to be converted object properties, usefull when converting a subset of the keys.
+     * Default is '', representing the root of the object.
+     */
+    prefix?: string;
+
+    /**
+     * The behavior when error or amibiguity occurs on converting hierarchical keys.
+     * Default is 'error'.
+     */
+    onError?: "error" | "ignore";
 }
 
 interface IGettable {
+    /**
+     * Get the value of a key-value from the Map-styled data structure.
+     * @param key - The key of the key-value to be retrieved.
+     */
     get<T>(key: string): T | undefined;
 }
--- a/src/AzureAppConfiguration.ts
+++ b/src/AzureAppConfiguration.ts
@@ -16,4 +16,12 @@ export type AzureAppConfiguration = {
      * @param thisArg - Optional. Value to use as `this` when executing callback.
      */
     onRefresh(listener: () => any, thisArg?: any): Disposable;
-} & ReadonlyMap<string, any>;
+} & IGettable & AppConfigurationData;
+
+interface AppConfigurationData {
+    data: { [key: string]: any };
+}
+
+interface IGettable {
+    get<T>(key: string): T | undefined;
+}
--- a/src/AzureAppConfiguration.ts
+++ b/src/AzureAppConfiguration.ts
@@ -29,7 +29,8 @@ interface IConfigurationObject {
 export interface ConfigurationObjectConstructionOptions {
     /**
      * The separator to use when converting hierarchical keys to object properties.
-     * Default is '.'.
+     * If separator is undefined, '.' will be used by default.
+
      */
     separator?: string;
 }
--- a/src/AzureAppConfigurationImpl.ts
+++ b/src/AzureAppConfigurationImpl.ts
@@ -184,6 +184,10 @@ export class AzureAppConfigurationImpl implements AzureAppConfiguration {
      */
     constructConfigurationObject(options?: ConfigurationObjectConstructionOptions): Record<string, any> {
         const separator = options?.separator ?? ".";
+        const validSeparators = [".", ",", ";", "-", "_", "__", "/", ":"];
+        if (!validSeparators.includes(separator)) {
+            throw new Error(`Invalid separator '${separator}'. Supported values: ${validSeparators.map(s => `'${s}'`).join(", ")}.`);
+        }
 
         // construct hierarchical data object from map
         const data: Record<string, any> = {};
--- a/src/AzureAppConfigurationImpl.ts
+++ b/src/AzureAppConfigurationImpl.ts
@@ -203,12 +203,17 @@ export class AzureAppConfigurationImpl implements AzureAppConfiguration {
                 }
                 // The path has been occupied by a non-object value, causing ambiguity.
                 if (typeof current[segment] !== "object") {
-                    throw new Error(`The key '${segments.slice(0, i + 1).join(separator)}' is not a valid path.`);
+                    throw new Error(`Ambiguity occurs when constructing configuration object from key '${key}', value '${value}'. The path '${segments.slice(0, i + 1).join(separator)}' has been occupied.`);
                 }
                 current = current[segment];
             }
+
+            const lastSegment = segments[segments.length - 1];
+            if (current[lastSegment] !== undefined) {
+                throw new Error(`Ambiguity occurs when constructing configuration object from key '${key}', value '${value}'. The key should not be part of another key.`);
+            }
             // set value to the last segment
-            current[segments[segments.length - 1]] = value;
+            current[lastSegment] = value;
         }
         return data;
     }
--- a/src/AzureAppConfigurationImpl.ts
+++ b/src/AzureAppConfigurationImpl.ts
@@ -3,7 +3,7 @@
 
 import { AppConfigurationClient, ConfigurationSetting, ConfigurationSettingId, GetConfigurationSettingOptions, GetConfigurationSettingResponse, ListConfigurationSettingsOptions } from "@azure/app-configuration";
 import { RestError } from "@azure/core-rest-pipeline";
-import { AzureAppConfiguration, HierarchicalDataConversionOptions } from "./AzureAppConfiguration";
+import { AzureAppConfiguration, ConfigurationObjectConstructionOptions } from "./AzureAppConfiguration";
 import { AzureAppConfigurationOptions } from "./AzureAppConfigurationOptions";
 import { IKeyValueAdapter } from "./IKeyValueAdapter";
 import { JsonKeyValueAdapter } from "./JsonKeyValueAdapter";
@@ -182,7 +182,7 @@ export class AzureAppConfigurationImpl implements AzureAppConfiguration {
     /**
      * Construct hierarchical data object from map.
      */
-    toHierarchicalData(options?: HierarchicalDataConversionOptions): Record<string, any> {
+    constructConfigurationObject(options?: ConfigurationObjectConstructionOptions): Record<string, any> {
         const separator = options?.separator ?? ".";
         const prefix = options?.prefix ?? "";
         const onError = options?.onError ?? "error";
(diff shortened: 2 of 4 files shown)


Eskibear:
Offline synced with Zhenlan, two improvements
-  AppConfig impl Map interfaces, and throw exception in set(k, v) to enforce immutability.

- Change `data` object to a helper function, e.g `constructDataObject(options)` (name TBD), which constructs an object using KVs from the configMap. The `options` can include:
  - customized `separator`, if necessary. default is `.` but might be `:` if the config is shared with other languages.
  - how to deal with ambiguity mentioned in above edge case, silently continue or throw exception.
  - customized `prefix`, default to empty string, meaning to convert the whole map. It's useful if a only configuration section is wanted.

src/AzureAppConfigurationImpl.ts
```diff
@@ -157,22 +174,51 @@ export class AzureAppConfigurationImpl extends Map<string, any> implements Azure
...
     }
 
+    /**
+     * Construct hierarchical data object from map.
+     *
+     * @remarks
+     * This method is used to construct hierarchical data object from map, and it is called after the initial load and each refresh.
+     */
+    #constructDataFromMap() {
+        // clean data object in-place
```
rossgrambo, line 201: Do we want to do this in place? I assumed we were going to Object.Freeze data, and then fully rebuild it when updating (to make it immutable like the map).

Eskibear: The purpose of in-place operation was to keep the appConfig.data reference unchanged, thus downstream consumer (e.g. feature manager) always get the updated value after a refresh, via appConfig.data.someKey.

Now after discussion, we don't expose a `data` property directly, instead we provide a helper function `constructData(options)` returning a data object as [mentioned](https://github.com/Azure/AppConfiguration-JavaScriptProvider/pull/49#issuecomment-1931196697). After a refresh, customers should update the reference of data object by themselves, with a proper onRefresh callback.
E.g.
```ts
// before adopting appconfig
config = JSON.parse(fs.readFileSync("config.json"))

console.log(config.server.port)

// after adopting appconfig
appConfig = await load(...)
config = appConfig.constructData()
appConfig.onRefresh(() => { config = appConfig.constructData() }) // the callback ensure the config is always up-to-date

console.log(config.server.port)
```

rossgrambo:
> AppConfig impl Map interfaces

I don't think there is a Map interface. Javascript added Map, and Javascript doesn't have interfaces. Looks like typescript didn't add an interface for it as far as I can tell.

src/AzureAppConfigurationImpl.ts
rossgrambo: I don't think this change solves much- except maybe helping guide user expectations. 

For the user, there is functionally no difference between:
```javascript
config = appConfig.constructData();
// and
config = appConfig.data;
```

The separator and prefix are a bonus, but they could be placed in the load function options (and probably should anyway so you don't repeat it everytime you call "constructData")
```javascript
appConfig = await load({
  "connection_string": "~~~",
  "separator": ":",
})
```
And this aligns with python's `load` funciton.

zhiyuanliang-ms:
For the edge case:

> Edge case (for data property only):
> Hierarchical key-value pairs with overlapped key prefix:
>   key: "app3.settings" => value: "placeholder"
>   key: "app3.settings.fontColor" => value: "yellow"

> How to re-construct data object from key-values?
> config.data.app3.settings = ? 

Since the default shape is map and user needs to call `constructDataObject(options)` to use data object, is this a valid case?
If the user sets two key "app3.settings" and "app3.settings.fontColor", will he use the object instead of map to get configuration?

zhiyuanliang-ms:
@rossgrambo There is [`Map`](https://github.com/microsoft/TypeScript/blob/main/src/lib/es2015.collection.d.ts#L1) [interface](https://microsoft.github.io/PowerBI-JavaScript/interfaces/_node_modules_typedoc_node_modules_typescript_lib_lib_es2015_collection_d_.map.html) in TypeScript. 

My question is why to use `Map` interface and throw exception when `set` is called. Currently, we are using `ReadOnlyMap` interface, this makes more sense to me.

src/AzureAppConfigurationImpl.ts
```diff
@@ -157,22 +174,51 @@ export class AzureAppConfigurationImpl extends Map<string, any> implements Azure
...
+        for (const key of Object.keys(this.data)) {
+            delete this.data[key];
+        }
+        // construct hierarchical data object from map
+        for (const [key, value] of this.#configMap) {
+            const segments = key.split(this.#SEP);
```
zhiyuanliang-ms, line 207: Should we handle the corner case when key = "a...b"

src/AzureAppConfigurationImpl.ts
Eskibear: It’s unlike python. In JS here the preferred way is to use get(key). In such use case, the key is full key matching app config store, and the customer doesn’t need to care about the separator when calling load(). The separator is only related to how we restore the hierarchy in the data object if we need one.

Moreover, about constructing data object, as I mentioned there might be edge case that ambiguity occurs and you have to drop some information or directly fail. It’s unreasonable to fail the load() as the key values are valid in app config store and the get(key) works as expected. With a function constructData(…), we can allow user to explicitly specify the behavior here and fail this function instead of the load(). It’s more reasonable and clear.

The prefix is for “configuration section” support. In node-config, there is a popular usage that we put configs of different components into an all-in-one config.json file, with component names as top level keys. Usage is like dbConfig=config.get<DbConfig>(“db”), where DbConfig defines the shape of the section. Then in app code, dbConfig.url, dbConfig.username etc are used. The prefix option here offers to only construct part of the loaded key values into a data object that customers do care about. And it’s also dedicated to the data object construction instead of the load().

src/AzureAppConfigurationImpl.ts
Eskibear: Good catch. In the case segments can be empty strings, which I think we should skip the construction for this key value. Or maybe honor the mentioned option (specifying behavior when ambiguity occurs) to silently skip or to throw error directly?

zhiyuanliang-ms: I vote for throwing error.

src/AzureAppConfiguration.ts
```diff
@@ -16,12 +16,40 @@ export type AzureAppConfiguration = {
...
+} & IGettable & IHierarchicalData;
 
-interface AppConfigurationData {
-    data: { [key: string]: any };
+interface IHierarchicalData {
+    /**
+     * Convert the Map-styled data structure to hierarchical object properties.
+     * @param options - The options to control the conversion behavior.
+     */
+    toHierarchicalData(options?: HierarchicalDataConversionOptions): Record<string, any>;
```
Eskibear, line 26: FAQ: What's `Record`?
`Record<K, V>` is a built-in TypeScript utility type, identical to `{ [key: K]: V }`, for convenience.
See https://www.typescriptlang.org/docs/handbook/utility-types.html#recordkeys-type

src/AzureAppConfigurationImpl.ts
Eskibear: impl: throw an error/ingore depending on value of  `options.onError`.

src/AzureAppConfiguration.ts
jimmyca15: Is it possible to have object in JS where property keys are not `string`? I'm not seeing the value of `Record<string, any>`. Perhaps if it was `Record<string, SomeType>` then yes. But with string and any it seems no better than normal object.

Eskibear: object also accepts keys of [Symbol](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Symbol) type, which is a unique primitive value with optional description. That's not the case here.

`Record<string, any>` is identical to `{ [key: string]: any }`, which is what we are using now. And yes, Record is more useful especially for limited values of keys. I'm ok with using the latter if Record introduces confusion.

jimmyca15: What would it be if it wasn't `Record<string, any>`?

Eskibear: I meant to use `{ [key: string]: any }` directly without introducing TypeScript built-in utility type Record, as people might not knowing what a "Record" is. Will that look better?

Eskibear:
As the design is finalized, this PR is ready for review now. Please take a look.

src/AzureAppConfiguration.ts
```diff
@@ -16,4 +16,28 @@ export type AzureAppConfiguration = {
...
+     * Construct configuration object based on Map-styled data structure and hierarchical keys.
+     * @param options - The options to control the conversion behavior.
+     */
+    constructConfigurationObject(options?: ConfigurationObjectConstructionOptions): Record<string, any>;
+}
+
+export interface ConfigurationObjectConstructionOptions {
+    /**
+     * The separator to use when converting hierarchical keys to object properties.
+     * Default is '.'.
```
zhiyuanliang-ms, line 32:
```suggestion
     * If separator is undefined, '.' will be used as the default value.

```

src/AzureAppConfigurationImpl.ts
```diff
@@ -168,11 +175,44 @@ export class AzureAppConfigurationImpl extends Map<string, any> implements Azure
...
+                if (!segment) {
+                    throw new Error(`invalid key: ${key}`);
+                }
+                // create path if not exist
+                if (current[segment] === undefined) {
+                    current[segment] = {};
+                }
+                // The path has been occupied by a non-object value, causing ambiguity.
+                if (typeof current[segment] !== "object") {
+                    throw new Error(`The key '${segments.slice(0, i + 1).join(separator)}' is not a valid path.`);
```
zhiyuanliang-ms, line 206: I think we can make some tweak on the error message here. Instead of saying it is not a valid path, we can say there is ambiguity.

Eskibear: How about this:
```ts
new Error(`There is ambiguity constructing configuration object from key '${key}', value '${value}'.`)
```

src/AzureAppConfigurationImpl.ts
```diff
@@ -168,11 +175,44 @@ export class AzureAppConfigurationImpl extends Map<string, any> implements Azure
...
+                // undefined or empty string
+                if (!segment) {
+                    throw new Error(`invalid key: ${key}`);
+                }
+                // create path if not exist
+                if (current[segment] === undefined) {
+                    current[segment] = {};
+                }
+                // The path has been occupied by a non-object value, causing ambiguity.
+                if (typeof current[segment] !== "object") {
```
zhiyuanliang-ms, line 205: It seems like the order of the keys in configMap is important. If there are two key-values: "app.settings.font.color" = "red" and "app.settings.font" = 1

If the key-value "app.settings.font.color" = "red" is handled first, "app.settings.font" = 1 will override it.

src/AzureAppConfigurationImpl.ts
```diff
@@ -168,11 +175,44 @@ export class AzureAppConfigurationImpl extends Map<string, any> implements Azure
...
+                    throw new Error(`The key '${segments.slice(0, i + 1).join(separator)}' is not a valid path.`);
+                }
+                current = current[segment];
+            }
+            // set value to the last segment
+            current[segments[segments.length - 1]] = value;
```
zhiyuanliang-ms, line 211: What will happen if there is a key "a.  ”?
data[a][" "] this can work. But for this method, I think our expected usage will be data.a.xxx

src/AzureAppConfigurationImpl.ts
Eskibear: Good catch! Before the data assignment, we should ensure that the property is not occupied, i.e. its value should be undefined. 
I'll fix that and add a test case for it.

src/AzureAppConfigurationImpl.ts
zhiyuanliang-ms: > How about this:
> 
> ```ts
> new Error(`There is ambiguity constructing configuration object from key '${key}', value '${value}'.`)
> ```

How about
``` ts
new Error(`There is ambiguity constructing configuration object from key '${key}', value '${value}'. The path '${segments.slice(0, i + 1).join(separator)}' has been occupied.`)
```

Eskibear: your suggestion is better, providing more info about the ambiguity. BTW I would change it to `Ambiguity occurs when constructing .... `. 

Then for key-values:
- "app.settings.font", 1
- "app.settings.font.color", "red"

the error would be like:
> Ambiguity occurs when constructing configuration object from key 'app.settings.font.color', value 'red'. The path 'app.settings.font' has been occupied.

However, if I switch the order as you mentioned, the ambiguity check would be at the end, there seems no **trivial** way to provide clue about the occupied path.
I.e. key-values:
- "app.settings.font.color", "red"
- "app.settings.font", 1
> Ambiguity occurs when constructing configuration object from key 'app.settings.font', value '1'.

Eskibear: In this case, I can add "The key should not be part of another key." And the error looks like:
> Ambiguity occurs when constructing configuration object from key 'app.settings.font', value '1'. The key should not be part of another key.

src/AzureAppConfigurationImpl.ts
zhiyuanliang-ms: Make sense.

src/AzureAppConfigurationImpl.ts
```diff
@@ -168,11 +175,49 @@ export class AzureAppConfigurationImpl extends Map<string, any> implements Azure
...
-
         // Mark all settings have loaded at startup.
         this.#isInitialLoadCompleted = true;
     }
 
+    /**
+     * Construct hierarchical data object from map.
+     */
+    constructConfigurationObject(options?: ConfigurationObjectConstructionOptions): Record<string, any> {
+        const separator = options?.separator ?? ".";
```
avanigupta, line 186: Should we have some restrictions on what is allowed as a separator?

src/AzureAppConfiguration.ts
```diff
@@ -16,4 +16,29 @@ export type AzureAppConfiguration = {
...
+
+interface IConfigurationObject {
+    /**
+     * Construct configuration object based on Map-styled data structure and hierarchical keys.
+     * @param options - The options to control the conversion behavior.
+     */
+    constructConfigurationObject(options?: ConfigurationObjectConstructionOptions): Record<string, any>;
+}
+
+export interface ConfigurationObjectConstructionOptions {
```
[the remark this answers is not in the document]
Eskibear: Yes, we keep the options for extensibility. If additional feature requested from customers, we don’t have to add more and more parameters and won’t break existing usage.

src/AzureAppConfigurationImpl.ts
```diff
@@ -15,7 +15,12 @@ import { CorrelationContextHeaderName } from "./requestTracing/constants";
 import { createCorrelationContextHeader, requestTracingEnabled } from "./requestTracing/utils";
 import { KeyFilter, LabelFilter, SettingSelector } from "./types";
 
-export class AzureAppConfigurationImpl extends Map<string, any> implements AzureAppConfiguration {
+export class AzureAppConfigurationImpl implements AzureAppConfiguration {
+    /**
+     * Hosting key-value pairs in the configuration store.
+     */
+    #configMap: Map<string, any> = new Map<string, any>();
```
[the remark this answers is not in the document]
Eskibear: From user’s perspective, yes, as we don’t provide a set method. But internally we will clear and update the map in place after a refresh.

src/AzureAppConfigurationImpl.ts
Eskibear: Good question. Do we allow an arbitrary character as separator in import/export? Or a string with multiple characters?

Eskibear: Just checked import/export in portal, the separator supports an allowlist: 
image

For the "local.json -> import and flatten in azconfig-> restore hierarchy js provider" use case, it makes sense to restrict the allowed values and make it consistent with portal. But is it possible that we add more allowed separators in the future? If so, I would rather not to add the restriction. Anyway users have to specify the separator by their own. And I don't see potential problem there.

avanigupta: We already have these restrictions in [CLI](https://learn.microsoft.com/en-us/cli/azure/appconfig/kv?view=azure-cli-latest#az-appconfig-kv-export), [portal and ADO pipeline tasks](https://msazure.visualstudio.com/Azure%20AppConfig/_git/SyncClient.TypeScript?path=/libraries/azure-app-configuration-sync/src/internal/constants.ts&version=GBmain&line=18&lineEnd=19&lineStartColumn=1&lineEndColumn=1&lineStyle=plain&_a=contents). If we ever add a new separator to the list, we'd have to make this change everywhere. I think it makes sense to follow the existing pattern from other clients and restrict the list of allowed separators.

Eskibear: Ok, then it makes total sense.
 I’ll add the restriction. I can change type of “separator” from string to allowed values to benefit TS developers, and also add the restriction in the same time.

Eskibear: restriction added via f4953da

avanigupta (approved):

zhiyuanliang-ms (approved):

PR #50, opened 2024-03-18
Merge main to release/preview/v1

Eskibear:


zhiyuanliang-ms (approved):

avanigupta (approved):

avanigupta:
Reminder to merge this PR with a merge commit since we should preserve all feature commits in the release branch.

Eskibear:
I just revisited our [public docs](https://learn.microsoft.com/en-us/azure/azure-app-configuration/quickstart-javascript-provider?tabs=linux), and find there is usage of `has(key)` in sample code, which I think is valid.
But other public method of ReadonlyMap was removed when I changed it to impl IGettable. That would also break the use case. So, I would like to change it back to impl ReadonlyMap which is compatible with interface IGettable. And it would not affect what we've done for FM lib.

PR #52, opened 2024-03-20
Add back ReadonlyMap APIs

Eskibear:
See https://github.com/Azure/AppConfiguration-JavaScriptProvider/pull/50#issuecomment-2008652167

In previous refactoring, we changed ReadonlyMap to IGettable, with Map APIs like `get(key)` being removed. This PR adds them back, adding some test cases BTW.

### DIFF
--- a/src/AzureAppConfigurationImpl.ts
+++ b/src/AzureAppConfigurationImpl.ts
@@ -91,8 +91,37 @@ export class AzureAppConfigurationImpl implements AzureAppConfiguration {
         this.#adapters.push(new JsonKeyValueAdapter());
     }
 
+    // ReadonlyMap APIs
     get<T>(key: string): T | undefined {
         return this.#configMap.get(key);
+    }
+
+    forEach(callbackfn: (value: any, key: string, map: ReadonlyMap<string, any>) => void, thisArg?: any): void {
+        this.#configMap.forEach(callbackfn, thisArg);
+    }
+
+    has(key: string): boolean {
+        return this.#configMap.has(key);
+    }
+
+    get size(): number {
+        return this.#configMap.size;
+    }
+
+    entries(): IterableIterator<[string, any]> {
+        return this.#configMap.entries();
+    }
+
+    keys(): IterableIterator<string> {
+        return this.#configMap.keys();
+    }
+
+    values(): IterableIterator<any> {
+        return this.#configMap.values();
+    }
+
+    [Symbol.iterator](): IterableIterator<[string, any]> {
+        return this.#configMap[Symbol.iterator]();
     }
 
     get #refreshEnabled(): boolean {
--- a/test/load.test.ts
+++ b/test/load.test.ts
@@ -127,6 +127,40 @@ describe("load", function () {
         expect(settings.get("app.settings.fontFamily")).undefined;
     });
 
+    it("should also work with other ReadonlyMap APIs", async () => {
+        const connectionString = createMockedConnectionString();
+        const settings = await load(connectionString, {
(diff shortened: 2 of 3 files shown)


test/load.test.ts
```diff
@@ -110,6 +110,40 @@ describe("load", function () {
...
+        const settings = await load(connectionString, {
+            selectors: [{
+                keyFilter: "app.settings.*",
+                labelFilter: "\0"
+            }]
+        });
+        expect(settings).not.undefined;
+        expect(settings.has("app.settings.fontColor")).true;
+        expect(settings.has("app.settings.fontSize")).true;
+        expect(settings.has("app.settings.fontFamily")).false;
```
zhiyuanliang-ms, line 124: Let's add testcases for `size`, `entries`, `keys`, `values` and iterator as well.

Eskibear: added in 007c521

src/AzureAppConfigurationImpl.ts
```diff
@@ -91,10 +91,39 @@ export class AzureAppConfigurationImpl implements AzureAppConfiguration {
...
+
+    has(key: string): boolean {
+        return this.#configMap.has(key);
+    }
+
+    get size(): number {
+        return this.#configMap.size;
+    }
+
+    entries(): IterableIterator<[string, any]> {
```
avanigupta, line 111: I was looking at [this documentation](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Classes/extends#avoiding_inheritance), and they are defining some of these methods as generator functions. I'm guessing that's not useful in our scenario because we don't expect any async operations here?

Eskibear: In short, not related to sync or async. The impl of APIs are designed to return exactly the same as a Map does. BTW for Map, all above APIs are sync operations except `forEach`. 

The MDN implementation is just a demonstration of pros not using inheritance in some cases. The `ReadOnlyMap` impl mentioned is not standard but provides correct results via expected usage. It's like we create ReadOnlyMap.keys as a generator, in terms of what to yield, it delegates to the Map.keys() iterator. But the type is different from Map.keys. 
Below comparison should be clear enough.
```console
// paste ReadOnlyMap impl from MDN
> a = new Map()
Map(0) {}
> a.set("k", "v")
Map(1) { 'k' => 'v' }
> ak=a.keys()
[Map Iterator] { 'k' }
> ak.next()
{ value: 'k', done: false }
> ak.next()
{ value: undefined, done: true }

> b = new ReadOnlyMap([["k", "v"]])
ReadOnlyMap {}
> bk = b.keys()
Object [Generator] {}   // <--- different type and toString() impl
> bk.next()
{ value: 'k', done: false }  // <--- correctly yield 'k'
> bk.next()
{ value: undefined, done: true }
```

My ReadonlyMap implementation here doesn't directly get generator involved. It simply returns exactly the same `IterableIterator` type directly.

If you are interested in the concepts, see https://developer.mozilla.org/en-US/docs/Web/JavaScript/Guide/Iterators_and_Generators for reference about iterator/generator/iterable...

As for the async usage you mentioned, that's something like AsyncIterableIterator, which implements a different symbol [Symbol.asyncIterator](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Symbol/asyncIterator) rather than [Symbol.iterator](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Symbol/iterator)

avanigupta (approved):

PR #53, opened 2024-03-20
Add example for configuration object usage

Eskibear:


examples/configObject.mjs
```diff
@@ -0,0 +1,45 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT license.
+
+import * as dotenv from "dotenv";
+dotenv.config()
+
+/**
+ * This example demostrates how to construct a configuration object from settings loaded from Azure App Configuration.
```
avanigupta, line 8:
```suggestion
 * This example demonstrates how to construct a configuration object from settings loaded from Azure App Configuration.

```

examples/configObject.mjs
```diff
@@ -0,0 +1,45 @@
...
+// Licensed under the MIT license.
+
+import * as dotenv from "dotenv";
+dotenv.config()
+
+/**
+ * This example demostrates how to construct a configuration object from settings loaded from Azure App Configuration.
+ * If you are using configuration object instead of Map-styled settings, it would minimize the code changes required to use Azure App Configuration in your application.
+ *
+ * When you import configuration into Azure App Configuration from a local .json file, the keys are automatically flattened with specified separator (default is ".").
```
avanigupta, line 11: >  (default is ".")

We dont have any default separator for import

avanigupta (approved):

examples/configObject.mjs
Eskibear: fixed in dc4578d

examples/configObject.mjs
Eskibear: The typo is fixed in dc4578d

PR #55, opened 2024-04-08
Exclude feature flags from loaded settings

Eskibear:
#### Context information

According to Feature Management design, the loaded key-values should be:
- FM is enabled: feature flags are included as a whole, under "feature_management" key.
- FM is not enabled: feature flags are excluded. 

#### Current behavior

Feature flags will be loaded and treated as JSON objects if customer specifies proper selectors, e.g. `keyFilter: "
.appconfig.featureflag/*"`

#### Expected behavior

Before FM is supported, the default behavior should be consistent with FM is disabled. We should exclude feature flags. 
**This will avoid future breaking change.**

### DIFF
--- a/test/load.test.ts
+++ b/test/load.test.ts
@@ -62,6 +62,17 @@ const mockedKVs = [{
 }, {
     key: "app5.settings",
     value: "placeholder"
+}, {
+    key: ".appconfig.featureflag/Beta",
+    value: JSON.stringify({
+        "id": "Beta",
+        "description": "",
+        "enabled": true,
+        "conditions": {
+            "client_filters": []
+        }
+    }),
+    contentType: "application/vnd.microsoft.appconfig.ff+json;charset=utf-8"
 }
 ].map(createMockedKeyValue);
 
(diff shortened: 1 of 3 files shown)


zhiyuanliang-ms:
Will it be a separated PR to add feature flag option (the example usage in the design doc of JS FM)?

Eskibear:
@zhiyuanliang-ms Yes, there will be a separate PR supporting FM. That will not be included in 1.0.0-preview.

zhiyuanliang-ms (approved):

test/load.test.ts
```diff
@@ -62,6 +62,17 @@ const mockedKVs = [{
 }, {
     key: "app5.settings",
     value: "placeholder"
+}, {
+    key: ".appconfig.featureflag/Beta",
```
zhiyuanliang-ms, line 66: Can we avoid hard coding feature prefix and content type here?

https://github.com/Azure/azure-sdk-for-js/blob/main/sdk/appconfiguration/app-configuration/src/featureFlag.ts#L11

Eskibear: I don't think there's much difference.

To be serious, my preference is hard code, as it's a test case. It's clearer to see the raw input and output. And imagine that one day if there is a regression in SDK providing the wrong prefix/contentType, we can detect it.

PR #61, opened 2024-04-23
Add API tryRefresh

Eskibear:


Eskibear:
Hold on, think about whether we need this new `tryRefresh` API if it only swallows all exception, as customers can simply write `refresh().catch(/* do nothing */)` on their own without breaking the application potentially.
I notice in .NET provider, it basically catches all kinds of exceptions (but throws [AggregationException](https://github.com/Azure/AppConfiguration-DotnetProvider/blob/a9eacbc1ba62e12eb1a42c2a2ebb326289f5e0f8/src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationProvider.cs#L490-L491) in some cases), not sure in which cases we want to throw exceptions.

PR #65, opened 2024-05-27
support feature flags

Eskibear:
By default feature flags are excluded and cannot be loaded. Only when `featureFlagOptions.enabled` is true, it loads feature flags as a whole into a preserved key `feature_management`, according to `featureFlagOptions.selectors`. The refresh behavior is independently controlled by `featureFlagOptions.refresh`.


src/AzureAppConfigurationImpl.ts
```diff
@@ -1,7 +1,7 @@
 // Copyright (c) Microsoft Corporation.
 // Licensed under the MIT license.
 
-import { AppConfigurationClient, ConfigurationSetting, ConfigurationSettingId, GetConfigurationSettingOptions, GetConfigurationSettingResponse, ListConfigurationSettingsOptions, isFeatureFlag } from "@azure/app-configuration";
+import { AppConfigurationClient, ConfigurationSetting, ConfigurationSettingId, FeatureFlagValue, GetConfigurationSettingOptions, GetConfigurationSettingResponse, ListConfigurationSettingsOptions, featureFlagPrefix, isFeatureFlag, parseFeatureFlag } from "@azure/app-configuration";
```
avanigupta, line 4: This is coming from JS SDK but I'm surprised to see some of these in pascal case and others in camel case. Do you know why that's the case?

src/AzureAppConfigurationImpl.ts
```diff
@@ -14,6 +14,7 @@ import { RefreshTimer } from "./refresh/RefreshTimer";
 import { CorrelationContextHeaderName } from "./requestTracing/constants";
 import { createCorrelationContextHeader, requestTracingEnabled } from "./requestTracing/utils";
 import { KeyFilter, LabelFilter, SettingSelector } from "./types";
+import { featureFlagsKeyName, featureManagementKeyName } from "./featureManagement/constants";
```
avanigupta, line 17: Can we make the casing consistent with other constants? For example, CorrelationContextHeaderName is also a constant, but the casing is different.

src/RefreshOptions.ts
```diff
@@ -25,3 +25,17 @@ export interface RefreshOptions {
      */
     watchedSettings?: WatchedSetting[];
 }
+
+export interface FeatureFlagRefreshOptions {
+    /**
+     * Specifies whether the provider should automatically refresh when the configuration is changed.
```
avanigupta, line 31:
```suggestion
     * Specifies whether the provider should automatically refresh all configurations if any feature flag changes.

```

src/RefreshOptions.ts
```diff
@@ -25,3 +25,17 @@ export interface RefreshOptions {
...
+
+export interface FeatureFlagRefreshOptions {
+    /**
+     * Specifies whether the provider should automatically refresh when the configuration is changed.
+     */
+    enabled: boolean;
+
+    /**
+     * Specifies the minimum time that must elapse before checking the server for any new changes.
+     * Default value is 10 seconds. Must be greater than 1 second.
```
avanigupta, line 37: Default should be 30 seconds, right?

src/RefreshOptions.ts
avanigupta: Just to confirm - we're not going to refresh key-values with FeatureFlag changes and FF with kv changes, right? Both refreshes are done independently of each other?

src/featureManagement/FeatureFlagOptions.ts
```diff
@@ -0,0 +1,29 @@
...
+     * Specifies whether feature flag support is enabled.
+     */
+    enabled: boolean;
+
+    /**
+     * Specifies the selectors used to filter feature flags.
+     *
+     * @remarks
+     * keyFilter of selector will be prefixed with "appconfig.featureflag/" when request is sent.
+     * If no selectors are specified then no feature flags will be retrieved.
```
avanigupta, line 21: If no selectors are specified, shouldn't we load all Feature Flags with null label? That's what we do for key-values too.

src/featureManagement/FeatureFlagOptions.ts
```diff
@@ -0,0 +1,29 @@
...
+
+import { FeatureFlagRefreshOptions } from "../RefreshOptions";
+import { SettingSelector } from "../types";
+
+/**
+ * Options used to configure feature flags.
+ */
+export interface FeatureFlagOptions {
+    /**
+     * Specifies whether feature flag support is enabled.
```
avanigupta, line 12:
```suggestion
     * Specifies whether feature flags will be loaded from Azure App Configuration.

```

src/AzureAppConfigurationImpl.ts
```diff
@@ -195,17 +223,58 @@ export class AzureAppConfigurationImpl implements AzureAppConfiguration {
...
+            };
+            if (this.#requestTracingEnabled) {
+                listOptions.requestOptions = {
+                    customHeaders: {
+                        [CorrelationContextHeaderName]: createCorrelationContextHeader(this.#options, this.#isInitialLoadCompleted)
+                    }
+                }
+            }
+
+            const settings = this.#client.listConfigurationSettings(listOptions);
```
avanigupta, line 256: Did JS SDK release collection monitoring API? I see they had a PR for this but not sure if this exists in the latest stable release: https://github.com/Azure/azure-sdk-for-js/pull/28555

src/AzureAppConfigurationImpl.ts
```diff
@@ -298,9 +376,25 @@ export class AzureAppConfigurationImpl implements AzureAppConfiguration {
...
     }
 
+    async #refreshFeatureFlags(): Promise<void> {
+        // if still within refresh interval/backoff, return
+        if (!this.#featureFlagRefreshTimer.canRefresh()) {
+            return Promise.resolve();
+        }
+
+        try {
+            await this.#loadFeatureFlags();
```
avanigupta, line 386: we should check if any feature flags changed before loading all feature flags again. I think collection monitoring API can be leveraged here.

src/AzureAppConfigurationImpl.ts
Eskibear: In JS the naming convention is usually as below:
- CONSTANT_VARIABLE_NAME
- TypeName
- variableName
- methodName 

In JS SDK they regard constants as variables and use constantVariableName. Makes sense to some extent.
In this repo, some constants are named as ConstantVariableName when copied from dotnet for reference, which I think should eventually be changed.

src/RefreshOptions.ts
Eskibear: Kv is 30s, this 10s is referred from k8s provide. I shall double confirm that. So in dotnet it’s 30s right?

src/featureManagement/FeatureFlagOptions.ts
Eskibear: Per discussion, for feature flags we expect customers to explicitly specify selectors.

src/AzureAppConfigurationImpl.ts
Eskibear: Agree. Ultimately we should leverage the collection monitoring api. I prefer we simply poll all ffs here for simplicity. I’ll check if it’s ready in latest js sdk and impl that in a separate PR.

src/AzureAppConfigurationImpl.ts
Eskibear: The PR is in sdk v1.6.0. Later I'll create another PR to update the dependencies and support the page etag.

src/AzureAppConfigurationImpl.ts
Eskibear: Sure, that's also the thing I want to do. Constants should be in`UPPER_SNAKE_CASE`, but in SDK it's `camelCase` starting with lowercase which is the same as normal variable names. 

Anyway, I prefer to use UPPER_SNAKE_CASE honoring the convention. I'll do that in another PR.

src/AzureAppConfigurationImpl.ts
Eskibear: unify constant naming in #68

src/AzureAppConfigurationImpl.ts
Eskibear: unify constant naming in https://github.com/Azure/AppConfiguration-JavaScriptProvider/pull/68

src/RefreshOptions.ts
Eskibear: updated in e269596

src/AzureAppConfigurationImpl.ts
Eskibear: merged from main branch, fixing the naming issue. Now other constants are all UPPER_SNAKE_CASE

src/AzureAppConfigurationImpl.ts
Eskibear: constants in this PR renamed in 708f759

src/AzureAppConfigurationImpl.ts
```diff
@@ -299,9 +377,25 @@ export class AzureAppConfigurationImpl implements AzureAppConfiguration {
...
+            this.#featureFlagRefreshTimer.reset();
+        } catch (error) {
+            // if refresh failed, backoff
+            this.#featureFlagRefreshTimer.backoff();
+            throw error;
+        }
+    }
```
avanigupta, line 394: Do we need to run callbacks after successful refresh of feature flags

src/AzureAppConfigurationImpl.ts
avanigupta: This could lead to a lot of server requests if someone is using a short refresh interval. It's almost like a forced refresh of feature flags whenever refresh interval elapses. This is conceptually very different from what other providers are doing. In other providers, we check the etag of each feature flag and reload feature flags only if at least 1 flag has changed on the server. Here's the Python implementation for reference: https://github.com/Azure/azure-sdk-for-python/blob/e08b3b0f8843a38407515b91a72638aa3dd09447/sdk/appconfiguration/azure-appconfiguration-provider/azure/appconfiguration/provider/_azureappconfigurationprovider.py#L584-L639

If the new Page etag API is close to being released, I would say lets implement that directly instead of tracking individual etags ourselves. Other providers were released before we had the API, so they implment the workaround of tracking etags manually. But either way, we should not reload feature flags if nothing has changed on the server.

src/featureManagement/FeatureFlagOptions.ts
avanigupta: Should we throw an error if `FeatureFlagOptions.enabled` is true but selectors are not provided?

src/AzureAppConfigurationImpl.ts
```diff
@@ -196,17 +224,58 @@ export class AzureAppConfigurationImpl implements AzureAppConfiguration {
...
+        for(const key of this.#configMap.keys()) {
+            if (key !== FEATURE_MANAGEMENT_KEY_NAME) {
+                this.#configMap.delete(key);
+            }
+        }
+    }
+
+    async #loadFeatureFlags() {
+        const featureFlags: FeatureFlagValue[] = [];
+        const featureFlagSelectors = getValidSelectors(this.#options?.featureFlagOptions?.selectors);
```
avanigupta, line 243: This will load all feature flags with null label by default - contrary to what the comment says for `FeatureFlagOptions.selectors`

src/AzureAppConfigurationImpl.ts
avanigupta: Is this version released now? I can see the public docs for the new API https://learn.microsoft.com/en-us/javascript/api/@azure/app-configuration/listconfigurationsettingpage?view=azure-node-latest

src/AzureAppConfigurationImpl.ts
```diff
@@ -196,17 +224,58 @@ export class AzureAppConfigurationImpl implements AzureAppConfiguration {
...
+                appConfigOptions: this.#options
+            };
+            const settings = listConfigurationSettingsWithTrace(
+                requestTraceOptions,
+                this.#client,
+                listOptions
+            );
+            for await (const setting of settings) {
+                if (isFeatureFlag(setting)) {
+                    const flag = parseFeatureFlag(setting);
```
avanigupta, line 261: What do you think about supporting V2 schema for feature flags as defined here: https://github.com/Azure/AppConfiguration/blob/main/docs/FeatureManagement/FeatureFlag.v2.0.0.schema.json ?

`parseFeatureFlag` returns [FeatureFlagValue](https://learn.microsoft.com/en-us/javascript/api/@azure/app-configuration/featureflagvalue?view=azure-node-latest#@azure-app-configuration-featureflagvalue-conditions) object, which does not have all the fields that we need.

src/AzureAppConfigurationImpl.ts
Eskibear: It's already released in js-sdk v1.6.0. Sounds reasonable, we should make sure it's correctly implemented before we ship it... I'll work on the corresponding impl.

src/AzureAppConfigurationImpl.ts
Eskibear: no we don't, as we don't have an similar `onRefresh()` api compared with configuration setting refresh.

src/featureManagement/FeatureFlagOptions.ts
Eskibear: I agree, as it makes no sense to enable feature flags but load nothing. Throwing an error is better than silently loading nothing.

src/AzureAppConfigurationImpl.ts
Eskibear: Supporting FeatureFlag.v2 schema is on the plan, but is not included in this PR. 
BTW I think it's better to update SDK impl first, as in the latest SDK 1.6.0 the [impl ](https://unpkg.com/@azure/app-configuration@1.6.0/dist/index.js)https://[unpkg.com/@azure/app-configuration@1.6.0/dist/index.js](https://unpkg.com/@azure/app-configuration@1.6.0/dist/index.js)hasn't cover variants yet.

src/featureManagement/FeatureFlagOptions.ts
Eskibear: 923c124

src/AzureAppConfigurationImpl.ts
Eskibear: fixed in 923c124

src/AzureAppConfigurationImpl.ts
avanigupta: Should we add that API? How can we enable users to run some custom code when feature flag refresh succeeds?

Eskibear: > Should we add that API? 

No idea, depends on whether there is valid custom requests.
Since `refresh()` returns a promise, one can always use `refresh().then(successCallback).catch(errorCallback)` to run custom code.
Currently public APIs `refresh()` is shared between ff & kv, and `onRefresh()` is dedicated to kv (based on current descrption). A little bit inconsistent as you mentioned. 

 So two options in my mind, I prefer a) for simplicity and not over-optimizing it.
- a) `refresh()` and `onRefresh()` both shared between ff&kv. I.e. we run callbacks after successfully refresh **BOTH** kv and ff.
- b) add a new API like `onFeatureFlagRefresh()`, (and a new `refreshFeatureFlags()` which is consistent but less easy to use) if we want to focus on ff refresh, and run callbacks even if kv refresh fails. It sounds too higher granularity, and I don't know if that's necessary.



For reference, just curious about what's the case in dotnet provider.

src/AzureAppConfigurationImpl.ts
avanigupta: SDK updates are usually slow so we dont rely on SDK for schema definitions. We parse the raw value manually in other providers too (example: [Python provider](https://github.com/Azure/azure-sdk-for-python/blob/66d5de409d0980d3b2c7b82971cc648f9d5cd6a5/sdk/appconfiguration/azure-appconfiguration-provider/azure/appconfiguration/provider/_azureappconfigurationprovider.py#L686)).

If we're adding the support for feature flags now, I'd suggest going straight for V2 schema since that's the one we want users to move to eventually.

src/AzureAppConfigurationImpl.ts
avanigupta: We dont provide any special way to register config change callbacks, but .NET has a way for users to implement [options pattern](https://learn.microsoft.com/en-us/dotnet/core/extensions/options#ioptionsmonitor) to get notified of configuration changes. This is triggered every time any config value changes (regardless of kv of ff). So I like your first option too. BTW, is there a way for users to tell what was refreshed - kv or ff?

Eskibear: > , is there a way for users to tell what was refreshed - kv or ff?

So far, for option a), no. The experience would be: you provide custom callbacks either via `refresh().then()` or `onRefresh()`, and after `refresh()` is successful (meaning both ff and kv are refreshed), the callback is executed, without knowing which kv of ff is actually updated. 
The design was because you can always get the up-to-date kv/ff by your own in the callback, and we don't compare the new/old values for you. If you are interested in the changed ones, you can always keep a cache and compare by your own.

src/AzureAppConfigurationImpl.ts
Eskibear: Makes sense. Since we now unified the schema, the config provider is just responsible for providing feature flags. 

BTW after revisiting SDK's impl, I also find another reason that we should not use `FeatureFlagValue` and `parseFeatureFlag`. 
In type `FeatureFlagValue` the properties are renamed from the original schema, following common JS naming convention. e.g. client_filters => clientFilters...etc.

Eskibear: now feature flags are parsed directly, v2 schema supported.  8e2de54

src/AzureAppConfigurationImpl.ts
avanigupta: That's reasonable. We should also plan to add logging statements to provide details of refresh events in future. For this PR, option A sounds good.

src/AzureAppConfigurationImpl.ts
```diff
@@ -371,12 +465,7 @@ export class AzureAppConfigurationImpl implements AzureAppConfiguration {
     }
 }
 
-function getValidSelectors(selectors?: SettingSelector[]) {
-    if (!selectors || selectors.length === 0) {
-        // Default selector: key: *, label: \0
-        return [{ keyFilter: KeyFilter.Any, labelFilter: LabelFilter.Null }];
-    }
-
+function validateSelectors(selectors: SettingSelector[]) {
```
avanigupta, line 468: I liked the name `getValidSelectors` for this method because we are returning selectors, not just validating them.

Eskibear: name updated with return type declaration added. 1182638

src/AzureAppConfigurationImpl.ts
Eskibear: > add logging statements to provide details of refresh events in future

tracked in #75

Eskibear: 99f2c29

`refresh()` and `onRefresh()` now apply for both key-values and feature flags. `refresh()` will try refreshing both kv and ff. If either kv of ff is successfully refreshed, the callbacks registered via `onRefresh()` will be called.

Pending task is to refresh feature flags using page etag.

src/AzureAppConfigurationImpl.ts
Eskibear: update sdk version to 1.6 in #76 . and we update this PR to use page etag api here then.

src/AzureAppConfigurationImpl.ts
```diff
@@ -45,6 +52,9 @@ export class AzureAppConfigurationImpl implements AzureAppConfiguration {
     #featureFlagRefreshInterval: number = DEFAULT_REFRESH_INTERVAL_IN_MS;
     #featureFlagRefreshTimer: RefreshTimer;
 
+    // selectors
+    #featureFlagSelectors: PagedSettingSelector[] = [];
```
Eskibear, line 56: Note: Here for each selector, both etag and corresponding items of each page are cached. The benefit is, when one of the pages is changed, we can get items of unchanged pages (without sending extra requests when 304) and do the composition among selectors.

But, this leads to more memory consumption. Since configuration/feature flags don't change very often,  in most requests it would be 304. Discussed with Zhenlan, and we can only cache the page etags without the items. When attempting to refresh, if all pages are 304, we do nothing. If any of the page is 200, we send requests again forcing to listing all the pages, and update the cached page etags. I will update the PR to go with this approach.

Eskibear: SDK v1.6.0 has a bug, when nothing changes, on listing configuration settings with pageEtags, only the first page.status_code is 304, all the other pages are 200.
Created an issue to track: https://github.com/Azure/azure-sdk-for-js/issues/30282

But the bug doesn't affects our impl in this PR, and actually not much impact, as in most of our use cases there should be only one page i believe. Just FYI.

src/AzureAppConfigurationImpl.ts
Eskibear: 4ec140a : refresh feature flags with page etag check first.. 

Providing page etags on listing configs, if all pages from all selectors return 304, then there is no need to refresh. 
If any page returns 200, we short-circuit the loop and load all feature flags from scratch, do the composition, and update pageEtags.

src/AzureAppConfigurationImpl.ts
Eskibear: SDK v1.6.1 is out carrying the fix.

src/AzureAppConfigurationImpl.ts
```diff
@@ -196,17 +240,65 @@ export class AzureAppConfigurationImpl implements AzureAppConfiguration {
...
+                    if (isFeatureFlag(setting)) {
+                        featureFlagsMap.set(setting.key, setting.value);
+                    }
+                }
+            }
+            selector.pageEtags = pageEtags;
+        }
+
+        // parse feature flags
+        const featureFlags = Array.from(featureFlagsMap.values()).map(rawFlag => JSON.parse(rawFlag));
```
avanigupta, line 288: What's the behavior if there are failures while parsing FF due to invalid JSON?

Eskibear: It would throw a SyntaxError (by JSON.parse), same as SDK [does](https://github.com/Azure/azure-sdk-for-js/blob/d328d966b33449e58b729ef1cc792fe307b64ff4/sdk/appconfiguration/app-configuration/src/featureFlag.ts#L102). And it fails the whole FF loading.

Invalid JSON in FF is a user error that we expect them to fix it first by themselves. I think it makes sense to simply throw the error and break the load.

src/AzureAppConfigurationImpl.ts
```diff
@@ -291,17 +417,73 @@ export class AzureAppConfigurationImpl implements AzureAppConfiguration {
...
+                    needRefresh = true;
+                    break;
+                }
+                // unchanged, check next page
+                pageCount++;
+            }
+
+            if (pageCount !== selector.pageEtags?.length) {
+                needRefresh = true; // page count changed indicating feature flags are added or deleted
+            }
```
avanigupta, line 462: This check is not needed. Service ensures that if pages are added or deleted, page etag will change, leading to 200 response. Thus, we don't need to maintain page count on client side.

src/AzureAppConfigurationImpl.ts
```diff
@@ -291,17 +417,73 @@ export class AzureAppConfigurationImpl implements AzureAppConfiguration {
...
+
+            if (needRefresh) {
+                break; // short-circuit if result from any of the selectors is changed
             }
         }
+
+        if (needRefresh) {
+            try {
+                await this.#loadFeatureFlags();
+                this.#featureFlagRefreshTimer.reset();
```
avanigupta, line 472: We should reset the refresh timer even if nothing changed on the server. This is to maintain `refresh interval` gap between refresh attempts. Same for regular kv.

src/AzureAppConfigurationImpl.ts
```diff
@@ -291,17 +417,73 @@ export class AzureAppConfigurationImpl implements AzureAppConfiguration {
...
+            return Promise.resolve(false);
+        }
+
+        // check if any feature flag is changed
+        let needRefresh = false;
+        for (const selector of this.#featureFlagSelectors) {
+            const listOptions: ListConfigurationSettingsOptions = {
+                keyFilter: `${featureFlagPrefix}${selector.keyFilter}`,
+                labelFilter: selector.labelFilter,
+                pageEtags: selector.pageEtags
```
avanigupta, line 442: How does the SDK behave when we pass empty (list of page etags (or null)? I'm imagining a scenario where the first request to load feature flags returned 0 results.

src/AzureAppConfigurationImpl.ts
```diff
@@ -196,17 +240,65 @@ export class AzureAppConfigurationImpl implements AzureAppConfiguration {
...
+            };
+
+            const pageEtags: string[] = [];
+            const pageIterator = listConfigurationSettingsWithTrace(
+                this.#requestTraceOptions,
+                this.#client,
+                listOptions
+            ).byPage();
+            for await (const page of pageIterator) {
+                if (page._response.status === 200) {
```
avanigupta, line 273: When we load FF without page etags, the expected response is always 200. Anything else will throw an exception.

src/AzureAppConfigurationImpl.ts
```diff
@@ -196,17 +240,65 @@ export class AzureAppConfigurationImpl implements AzureAppConfiguration {
...
+                    if (page.etag) {
```
avanigupta, line 274: This check covers null or empty both? If the etag is null or empty, I think we should still push it into the list of page etags. This should never happen, but if it does happen due to some server bug, we need to preserve the order of server returned page etags. Otherwise, the next request to load FF using this list of page etags would be indeterministic.

src/AzureAppConfigurationImpl.ts
```diff
@@ -196,17 +240,65 @@ export class AzureAppConfigurationImpl implements AzureAppConfiguration {
...
+                }
+            }
+            selector.pageEtags = pageEtags;
+        }
+
+        // parse feature flags
+        const featureFlags = Array.from(featureFlagsMap.values()).map(rawFlag => JSON.parse(rawFlag));
+
+        // feature_management is a reserved key, and feature_flags is an array of feature flags
+        this.#configMap.set(FEATURE_MANAGEMENT_KEY_NAME, { [FEATURE_FLAGS_KEY_NAME]: featureFlags });
```
avanigupta, line 291: Will this overwrite older feature flags in `configMap`?

src/AzureAppConfigurationImpl.ts
Eskibear: It makes sense to keep the order, I'll remove the check here, and the 200 status code check above.

More info:
In SDK, page.etag defintion as below, can be either `string` or `undefined`. 
It cannot be `null`, otherwise it would be a bug of SDK.

```ts
/**
 * Entity with etag.
 */
export declare interface EtagEntity {
    /**
     * The etag for this entity
     */
    etag?: string;
}
```

the `if(page.etag)` check is false if `page.etag` is undefined or null or empty string or even a number `0`. But given the type definition, the possible/legal values are string or undefined. So the check is identical to "isUndefinedOrEmptyString".

src/AzureAppConfigurationImpl.ts
Eskibear: then I'll simply remove the check.

src/AzureAppConfigurationImpl.ts
Eskibear: yes, it overwrites value with key "feature_management" in the configMap.

src/AzureAppConfigurationImpl.ts
Eskibear: > How does the SDK behave when we pass empty (list of page etags (or null)? 

SDK client will send requests without adding etag related headers. See:
https://github.com/Azure/azure-sdk-for-js/blob/6c4117cf31e82aedcf2ecffe079f1899abfbe5dd/sdk/appconfiguration/app-configuration/src/appConfigurationClient.ts#L341

> a scenario where the first request to load feature flags returned 0 results.

When it returns 0 results, the page etag is not empty. What would be the concern?

src/AzureAppConfigurationImpl.ts
Eskibear: A sample: 

<html xmlns:o="urn:schemas-microsoft-com:office:office"
xmlns:x="urn:schemas-microsoft-com:office:excel"
xmlns="http://www.w3.org/TR/REC-html40">

<head>

<meta name=ProgId content=Excel.Sheet>
<meta name=Generator content="Microsoft Excel 15">
<link id=Main-File rel=Main-File
href="file:///C:/Users/yanzh/AppData/Local/Temp/msohtmlclip1/01/clip.htm">
<link rel=File-List
href="file:///C:/Users/yanzh/AppData/Local/Temp/msohtmlclip1/01/clip_filelist.xml">
<style>

</style>
</head>

<body link="#467886" vlink="#96607D">


items | pages | etags
-- | -- | --
99 | 1 | vl5Cd2p4D1lCoNLt7hZhoC2i6fYNCbRQu9vMWI2-Hxw
100 | 2 | Rgl_lrYN54YE5ABHRondZJQMvTKtISxZtxiCewDkK30     00wWELzclWef22f82qcUDI7FbJweVOoKWZDA_JoUeD0
101 | 2 | Rgl_lrYN54YE5ABHRondZJQMvTKtISxZtxiCewDkK30     HYBJUIuXa4rJRH7dfYomEEz_RLiEpFqxhn81I3_aW38



</body>

</html>

Eskibear: > Service ensures that if pages are added or deleted, page etag will change,

Indeed. I'll remove the pageCount check.

src/AzureAppConfigurationImpl.ts
Eskibear: BTW, as SDK defines `pageEtags` as `string[]`, meaning each element of the pageEtags list can only be of `string` type. Thus we cannot put `undefined` as a placeholder. I would use an empty string for the case when pageEtag of the response is null or empty.

src/AzureAppConfigurationImpl.ts
Eskibear: fb56a0f

src/AzureAppConfigurationImpl.ts
Eskibear: b1cf809

src/AzureAppConfigurationImpl.ts
Eskibear: 3c12654

src/AzureAppConfigurationImpl.ts
Eskibear: 3c12654

src/AzureAppConfigurationImpl.ts
avanigupta: Ah I forgot that an empty page with 0 kv still has a valid etag.

avanigupta (approved):

src/featureManagement/FeatureFlagOptions.ts
zhiyuanliang-ms: #157

PR #79, opened 2024-07-04
parse JSON string resolved from kvr

### LINKED ISSUE #78: "application/json" content type is not handled in KeyVault references
Internally AzureKeyVaultKeyValueAdapter doesn't handle ContentType property of secret value and returns naked value. 

So we have dual behavior when same JSON setting stored directly or using keyvault reference with application/json content type returned as object or string. 

I think problem seats at this peace of AzureKeyVaultKeyValueAdapter  code 
`        if (client) {
            // TODO: what if error occurs when reading a key vault value? Now it breaks the whole load.
            const secret = await client.getSecret(secretName, { version });
            return [setting.key, secret.value]; // here it returns raw value without considering secret.properties.contentType value
        }
`

Thank you,

Eskibear:
fix #78 

Now if a JSON string is stored in Key Vault with content type "application/json", it will be loaded as parsed object instead of a raw string.

### DIFF
--- a/src/IKeyValueAdapter.ts
+++ b/src/IKeyValueAdapter.ts
@@ -5,12 +5,11 @@ import { ConfigurationSetting } from "@azure/app-configuration";
 export interface IKeyValueAdapter {
     /**
      * Determine whether the adapter applies to a configuration setting.
-     * Note: A setting is expected to be processed by at most one adapter.
      */
     canProcess(setting: ConfigurationSetting): boolean;
 
     /**
-     * This method process the original configuration setting, and returns processed key and value in an array.
+     * This method process the original configuration setting in place.
      */
-    processKeyValue(setting: ConfigurationSetting): Promise<[string, unknown]>;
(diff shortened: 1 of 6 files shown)


src/IKeyValueAdapter.ts
```diff
@@ -5,12 +5,11 @@ import { ConfigurationSetting } from "@azure/app-configuration";
 export interface IKeyValueAdapter {
     /**
      * Determine whether the adapter applies to a configuration setting.
-     * Note: A setting is expected to be processed by at most one adapter.
```
Eskibear, line 8: @avanigupta 
Here the use case is to store some correlated configs altogether into an object and JSON.stringify it. Maybe for some security reason, put it into a key vault, marking it as application/json. For JavaScript developers it looks a valid case. We should auto-parse it as we see the content type.

In dotnet, only one key value adapter would process a setting. But here to support the case, we should remove the constraints. The adapters should work like pipeline processors. We can rename it if you feel strange about the name "adapter".

Issue #81, opened 2024-07-25
WorkerNavigator is not defined


richard-thai:
Hi 

I'm encountering an error "WorkerNavigator is not defined" while testing @azure/app-configuration-provider on a NextJS 14 project using App Router. 

Any assistance would be appreciated. Thank you.

**stacktrace**
```
ReferenceError: WorkerNavigator is not defined
    at **isWebWorker** (webpack-internal:///(rsc)/../../node_modules/.pnpm/@azure+app-configuration-provider@1.0.0/node_modules/@azure/app-configuration-provider/dist-esm/requestTracing/utils.js:97:137)
    at getHostType (webpack-internal:///(rsc)/../../node_modules/.pnpm/@azure+app-configuration-provider@1.0.0/node_modules/@azure/app-configuration-provider/dist-esm/requestTracing/utils.js:74:14)
    at createCorrelationContextHeader (webpack-internal:///(rsc)/../../node_modules/.pnpm/@azure+app-configuration-provider@1.0.0/node_modules/@azure/app-configuration-provider/dist-esm/requestTracing/utils.js:20:72)
    at #loadSelectedKeyValues (webpack-internal:///(rsc)/../../node_modules/.pnpm/@azure+app-configuration-provider@1.0.0/node_modules/@azure/app-configuration-provider/dist-esm/AzureAppConfigurationImpl.js:129:197)
    at #loadSelectedAndWatchedKeyValues (webpack-internal:///(rsc)/../../node_modules/.pnpm/@azure+app-configuration-provider@1.0.0/node_modules/@azure/app-configuration-provider/dist-esm/AzureAppConfigurationImpl.js:169:65)
    at AzureAppConfigurationImpl.load (webpack-internal:///(rsc)/../../node_modules/.pnpm/@azure+app-configuration-provider@1.0.0/node_modules/@azure/app-configuration-provider/dist-esm/AzureAppConfigurationImpl.js:185:52)
    at load (webpack-internal:///(rsc)/../../node_modules/.pnpm/@azure+app-configuration-provider@1.0.0/node_modules/@azure/app-configuration-provider/dist-esm/load.js:53:32)
    at eval (webpack-internal:///(rsc)/../../packages/core/src/config/app-configuration.ts:21:96)
    at __webpack_require__.a (C:\Projects\Framework\ui\template\nextjs\internal-webapp-basic\apps\webui\.next\server\webpack-runtime.js:105:13)
[...]

PR #82, opened 2024-07-25
Fix reference error in navigator detection

Eskibear:
When `WorkerNavigator` is undefined, `navigator instanceof WorkerNavigator` throws reference error. The `typeof SomeDefinedType` is expected to be "function".


Eskibear:
fix #81

zhiyuanliang-ms (approved):

PR #83, opened 2024-07-25
Merge main to release/stable/v1

Eskibear:

PR #84, opened 2024-07-25
Update version to 1.0.1

Eskibear:
requires #83 

Mainly to ship a fix for #81, unblocking customers from using this lib in some Browser related environments.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,12 +1,12 @@
 {
   "name": "@azure/app-configuration-provider",
-  "version": "1.0.0",
+  "version": "1.0.1",
   "lockfileVersion": 2,
   "requires": true,
   "packages": {
     "": {
       "name": "@azure/app-configuration-provider",
-      "version": "1.0.0",
+      "version": "1.0.1",
       "license": "MIT",
       "dependencies": {
         "@azure/app-configuration": "^1.6.0",
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@azure/app-configuration-provider",
-  "version": "1.0.0",
+  "version": "1.0.1",
   "description": "The JavaScript configuration provider for Azure App Configuration",
   "main": "dist/index.js",
   "module": "./dist-esm/index.js",
--- a/src/version.ts
+++ b/src/version.ts
@@ -1,4 +1,4 @@
 // Copyright (c) Microsoft Corporation.
 // Licensed under the MIT license.
 
-export const VERSION = "1.0.0";
+export const VERSION = "1.0.1";


juniwang (approved):

avanigupta:
Should this PR go to main branch first, and then main can be merged to release? We normally dont merge code changes directly to release branch.

Eskibear:
@avanigupta 
current branching model is: changes go into main branch, and are merged to release branch. Only PRs related to official version bumping go directly into release branch. And for each release we do tagging. Benefit is the two branches are almost consistent (even though not strictly), and for local/dev build, they don’t use the same official version by default. The disadvantage is like here, occasionally there would be conflicts requiring manual resolving.

Your suggestion also makes sense, making the branches strictly consistent. I.e from what I understand,
- main: default branch where most changes go into. And we bump stable version here as well.
- release/stable/*: based on main, used for releasing.
- preview: if necessarily to maintain two major version at the same time, based on main but ahead of main with preview changes.
- release/preview/*: based on preview, for releasing preview versions.

i'm ok to switch to what you mentioned. A slight disadvantage is: in main and preview branch, if any developer checks out code, makes some changes, and build a local package to dev/test, it sends telemetry with an official version but the code is unpredicted. might affect statistics somethimes, but not a big problem though.

avanigupta:
>  build a local package to dev/test, it sends telemetry with an official version

Is there a way to identify local environment and change the version? In .NET, we can identify this and change the version accordingly [(link)](https://github.com/Azure/AppConfiguration-DotnetProvider/blob/0a439885e82d66b6b4dac02980aa4ade61e8a7ba/src/Microsoft.Extensions.Configuration.AzureAppConfiguration/Microsoft.Extensions.Configuration.AzureAppConfiguration.csproj#L40-L46)

What is the release branch strategy followed by JS SDK team?

Eskibear:
> Is there a way to identify local environment and change the version? 

Basically the version is determined by `version` in package.json. Of course in practice we have multiple ways to distinguish dev/prod env.

E.g. we can add more build scripts:
- `npm run build:dev` updates the version in package.json first before building package.
- `npm run build:prod` directly builds the package.

The problem is we cannot build a package with local version without changing the package.json file. That leads to inconvenience when we want to check in code, because pacakge.json is dirty with unwanted version change. You have to be careful not to include them. That's why I don't like the solution and want to leave packageJSON.version unchanged during building.

BTW, the version in CorrelationHeader is determined by version.ts, which should be identical to packageJSON.version. With separate scripts to build dev/prod packages, it's feasible to update the file as well. This is not a problem.

Eskibear:
> What is the release branch strategy followed by JS SDK team?

JS SDK repo simply has one protected branch `main`. They create PR against the main branch, and do releasing with main branch, and add proper tagging to specific commit.
The SDK repo include multiple projects owned by different teams, probably not the one we want to follow.

Eskibear:
@avanigupta
Later I can proceed to switch the branching model, as I mentioned [above](https://github.com/Azure/AppConfiguration-JavaScriptProvider/pull/84#issuecomment-2251536992). Aligning with our other repos is also important and requires less additional context information, making it easier to maintain.

The steps would be:
- merge release back into main, with a merge commit.
- create a PR to update dev-build scripts, which can build local dev version without eventually changing package.json.

PR #85, opened 2024-07-26
Merge release back to main

Eskibear:
To make the two branches "even". 

In the future the version bumping is also against main branch. And we don't directly commit into release branch.

NOTE: merge this PR with a merge commit.


avanigupta:
There are a lot of extra merge commits in release branch which don't need to go into main branch. I think it's okay to skip merging all these commits into main branch to make it even with release branch. Going forward, we can bump the version directly in main branch and then merge to release. IMO, it's okay if the branches are not even at this time.

PR #91, opened 2024-08-20
Update linting rule to enforce to use semicolon

zhiyuanliang-ms:


linglingye001:
Is there github action that can run the lint rule job?

linglingye001 (approved):

PR #98, opened 2024-09-14
Failover support

linglingye001:


.github/workflows/ci.yml
```diff
@@ -2,9 +2,9 @@ name: AppConfiguration-JavaScriptProvider CI
 
 on:
   push:
-    branches: [ "main" ]
+    branches: [ "main", "preview" ]
```
zhiyuanliang-ms, line 5: Can you separate this change in another pr?

src/ConfigurationClientManager.ts
```diff
@@ -0,0 +1,332 @@
...
+import { ConfigurationClientWrapper } from "./ConfigurationClientWrapper";
+import { TokenCredential } from "@azure/identity";
+import { AzureAppConfigurationOptions, MaxRetries, MaxRetryDelayInMs } from "./AzureAppConfigurationOptions";
+import { isFailoverableEnv } from "./requestTracing/utils";
+import * as RequestTracing from "./requestTracing/constants";
+
+const TCP_ORIGIN = "_origin._tcp";
+const ALT = "_alt";
+const TCP = "_tcp";
+const EndpointSection = "Endpoint";
```
Eskibear, line 14: unify naming for constants?   `TCP_ORIGIN` vs `EndpointSection` here.

zhiyuanliang-ms:
How about seperating this PR into two smaller PR for the convenience of reviewing. For example, the change of adding client manager can be a seperated PR and I can approve it quickly. And with client manager change merged, it won't block the following new feature like snapshot support.

src/AzureAppConfigurationImpl.ts
```diff
@@ -258,29 +294,42 @@ export class AzureAppConfigurationImpl implements AzureAppConfiguration {
...
-        const featureFlagSettings: ConfigurationSetting[] = [];
-        for (const selector of this.#featureFlagSelectors) {
-            const listOptions: ListConfigurationSettingsOptions = {
-                keyFilter: `${featureFlagPrefix}${selector.keyFilter}`,
-                labelFilter: selector.labelFilter
-            };
+        // Temporary map to store feature flags, key is the key of the setting, value is the raw value of the setting
+        const funcToExecute = async (client) => {
+            const featureFlagSettings: ConfigurationSetting[] = [];
+            const selectors = JSON.parse(
```
RichardChen820, line 300: Any special reason to do serialize and deserialize the ff selectors?

src/ConfigurationClientManager.ts
```diff
@@ -0,0 +1,332 @@
...
+    #staticClients: ConfigurationClientWrapper[];
+    #dynamicClients: ConfigurationClientWrapper[];
+    #lastFallbackClientRefreshTime: number = 0;
+    #lastFallbackClientRefreshAttempt: number = 0;
+
+    constructor (
+        connectionStringOrEndpoint?: string | URL,
+        credentialOrOptions?: TokenCredential | AzureAppConfigurationOptions,
+        appConfigOptions?: AzureAppConfigurationOptions
+    ) {
```
RichardChen820, line 45: Could we use constructor overload here? I know this capability of TS that can help flexibly pass arguments, but it looks too hacky to input different set of parameters.

src/ConfigurationClientManager.ts
```diff
@@ -0,0 +1,332 @@
...
+ * Query SRV records and return target hosts.
+ * @param {string} host - The host to query.
+ * @returns {Promise<string[]>} - A promise that resolves to an array of target hosts.
+ */
+async function querySrvTargetHost(host) {
+    const results: string[] = [];
+    let dns;
+
+    if (isFailoverableEnv()) {
+        dns = require("dns/promises");
```
RichardChen820, line 174: Does it mean each time of querying srv target host, it will attempt to load the module? Not sure whether it has any performance issue.

src/ConfigurationClientManager.ts
```diff
@@ -0,0 +1,332 @@
...
+    } else {
+        return results;
```
RichardChen820, line 176: We actually don't expect to execute this method if it's not a failoverable environment right? Then let's throw exception here, rather than return an empty list.

src/ConfigurationClientManager.ts
```diff
@@ -0,0 +1,332 @@
...
+    }
+
+    try {
+        // Look up SRV records for the origin host
+        const originRecords = await dns.resolveSrv(`${TCP_ORIGIN}.${host}`);
```
RichardChen820, line 181: If there any difference of the returning value between "SRV not exists" and "SRV exists but no records"

src/ConfigurationClientManager.ts
```diff
@@ -0,0 +1,332 @@
...
+            } catch (err) {
+                if (err.code === "ENOTFOUND") {
+                    break; // No more alternate records, exit loop
+                } else {
+                    throw new Error(`Failed to lookup alternate SRV records: ${err.message}`);
+                }
+            }
+        }
+    } catch (err) {
+        throw new Error(`Failed to lookup origin SRV records: ${err.message}`);
```
RichardChen820, line 218: If line213 fail, will the error be wrapped into "Failed to lookup origin SRV records: failed to lookup alternate SRV records:  xxxx"?

src/AzureAppConfigurationImpl.ts
linglingye001: I'm not sure if it's for pageEtag? @Eskibear

src/ConfigurationClientManager.ts
linglingye001: Do you mean there's no need to check the `originRecords` length?

src/ConfigurationClientManager.ts
linglingye001: If error occurs when query originRecords, it will not execute the altRecords query.

src/ConfigurationClientManager.ts
RichardChen820: I mean what if the _origin._tcp of origin store does not exists?

src/ConfigurationClientManager.ts
RichardChen820: I don't get you, you throw error in line213, and it might be watched by line217,

src/ConfigurationClientManager.ts
linglingye001: I see, we should check the err.code like altRecordsQuery. Btw, no originRecords means server side error right?

RichardChen820: I believe so

RichardChen820: It's not a sever side error, it may just because of a store does not have any replica been created

src/ConfigurationClientManager.ts
linglingye001: Ohhhhh yes, if failed inside, the error will be propagated. The error msg outside should be update.

src/AzureAppConfigurationImpl.ts
zhiyuanliang-ms: @linglingye I didn't find this code in the original implementation. What's the reason for this change?

The type of this.#featureFlagSelectors is `PagedSettingSelector[]` it is already an array. I didn't see any reason to stringfy it and then deserialize it.

src/ConfigurationClientManager.ts
```diff
@@ -0,0 +1,332 @@
...
+        } else {
+            throw new Error("A connection string or an endpoint with credential must be specified to create a client.");
+        }
+
+        this.#staticClients = [new ConfigurationClientWrapper(this.endpoint, staticClient)];
+        this.#validDomain = getValidDomain(this.endpoint);
+        this.isFailoverable = (options?.replicaDiscoveryEnabled ?? true) && isFailoverableEnv();
+    }
+
+    async getClients() {
```
zhiyuanliang-ms, line 75: please add return type for each function

src/AzureAppConfigurationImpl.ts
```diff
@@ -176,34 +176,70 @@ export class AzureAppConfigurationImpl implements AzureAppConfiguration {
...
+            this.#clientManager.refreshClients();
+            throw new Error("No client is available to connect to the target App Configuration store.");
+        }
 
-        // validate selectors
-        const selectors = getValidKeyValueSelectors(this.#options?.selectors);
+        for (const client of clients) {
+            let successful = false;
+            try {
+                const result = await funcToExecute(client.client);
```
zhiyuanliang-ms, line 194: for (const clientWrapper of clientWrappers)

client.client looks bad. clientWrapper.client is better.

src/ConfigurationClientManager.ts
zhiyuanliang-ms: There is no overload in javascript. The best we can do is something like [this](https://github.com/Azure/AppConfiguration-JavaScriptProvider/blob/main/src/load.ts#L28) But i am also no sure whether it is possible for constructor @Eskibear

src/AzureAppConfigurationImpl.ts
```diff
@@ -176,34 +176,70 @@ export class AzureAppConfigurationImpl implements AzureAppConfiguration {
...
             initialLoadCompleted: this.#isInitialLoadCompleted,
-            appConfigOptions: this.#options
+            appConfigOptions: this.#options,
+            isFailoverRequest: this.#isFailoverRequest
         };
     }
 
-    async #loadSelectedKeyValues(): Promise<ConfigurationSetting[]> {
-        const loadedSettings: ConfigurationSetting[] = [];
+    async #executeWithFailoverPolicy(funcToExecute) {
```
zhiyuanliang-ms, line 184: could add type for funcToExecute

src/ConfigurationClientManager.ts
```diff
@@ -0,0 +1,336 @@
...
+        return dateTime >= this.#lastFallbackClientRefreshAttempt + MinimalClientRefreshInterval
+            && (!this.#dynamicClients
+                || this.#dynamicClients.every(client => dateTime < client.backoffEndTime)
+                || dateTime >= this.#lastFallbackClientRefreshTime + FallbackClientRefreshExpireInterval);
+    }
+}
+
+/**
+ * Query SRV records and return target hosts.
+ * @param {string} host - The host to query.
```
zhiyuanliang-ms, line 166: move this to function signature

src/ConfigurationClientManager.ts
```diff
@@ -0,0 +1,317 @@
...
+import { ConfigurationClientWrapper } from "./ConfigurationClientWrapper";
+import { TokenCredential } from "@azure/identity";
+import { AzureAppConfigurationOptions, MaxRetries, MaxRetryDelayInMs } from "./AzureAppConfigurationOptions";
+import { isFailoverableEnv } from "./requestTracing/utils";
+import * as RequestTracing from "./requestTracing/constants";
+
+const TCP_ORIGIN = "_origin._tcp";
+const ALT = "_alt";
+const TCP = "_tcp";
+const Endpoint = "Endpoint";
```
zhiyuanliang-ms, line 14: Could you update to the existing naming convention: https://github.com/Azure/AppConfiguration-JavaScriptProvider/blob/preview/src/featureManagement/constants.ts to make it consistent.

package-lock.json
```diff
@@ -2524,6 +2524,7 @@
       "resolved": "https://registry.npmjs.org/micromatch/-/micromatch-4.0.8.tgz",
       "integrity": "sha512-PXwfBhYu0hBCPw8Dn0E+WDYb7af3dSLVWKi3HGv84IdF4TyFoC0ysxFd0Goxw7nSv4T/PzEJQxsYsEiFCKo2BA==",
       "dev": true,
+      "license": "MIT",
```
zhiyuanliang-ms, line 2527: Did npm make this change?

src/AzureAppConfigurationImpl.ts
```diff
@@ -176,34 +176,70 @@ export class AzureAppConfigurationImpl implements AzureAppConfiguration {
...
+            throw new Error("No client is available to connect to the target App Configuration store.");
+        }
 
-        // validate selectors
-        const selectors = getValidKeyValueSelectors(this.#options?.selectors);
+        for (const clientWrapper of clientWrappers) {
+            let successful = false;
+            try {
+                const result = await funcToExecute(clientWrapper.client);
+                this.#isFailoverRequest = false;
```
zhiyuanliang-ms, line 195: isFailoverRequest should be set to false before iterating all clients.

Let's say we call load() and then call refresh().

During load(), a failover happend, the first client failed and the second client failed. Now `#isFailoverRequest` is set to true.

Then, we call refresh(), we are going to send request to the second client, but this request will be tagged as failover request.

src/AzureAppConfigurationImpl.ts
```diff
@@ -176,34 +176,70 @@ export class AzureAppConfigurationImpl implements AzureAppConfiguration {
...
+            }
+        }
 
-            const settings = listConfigurationSettingsWithTrace(
-                this.#requestTraceOptions,
-                this.#client,
-                listOptions
-            );
+        this.#clientManager.refreshClients();
+        throw new Error("All app configuration clients failed to get settings.");
```
zhiyuanliang-ms, line 211: Inconsistent with the error message in line 188. Please unify lowercase or uppercase for app configuration.

src/AzureAppConfigurationImpl.ts
```diff
@@ -176,34 +176,70 @@ export class AzureAppConfigurationImpl implements AzureAppConfiguration {
...
+                throw error;
+            }
+        }
 
-            const settings = listConfigurationSettingsWithTrace(
-                this.#requestTraceOptions,
-                this.#client,
-                listOptions
-            );
+        this.#clientManager.refreshClients();
```
zhiyuanliang-ms, line 210: No need to refresh here.

src/AzureAppConfigurationImpl.ts
```diff
@@ -176,34 +176,70 @@ export class AzureAppConfigurationImpl implements AzureAppConfiguration {
...
     }
 
-    async #loadSelectedKeyValues(): Promise<ConfigurationSetting[]> {
-        const loadedSettings: ConfigurationSetting[] = [];
+    async #executeWithFailoverPolicy(funcToExecute) {
+        const clientWrappers = await this.#clientManager.getClients();
+        if (clientWrappers.length === 0) {
+            this.#clientManager.refreshClients();
+            throw new Error("No client is available to connect to the target App Configuration store.");
+        }
```
zhiyuanliang-ms, line 189: There are two refreshClients calls and two places to throw exception, however if you remove these lines, the behavior remains the same, except for the slight difference in error message.

src/AzureAppConfigurationImpl.ts
```diff
@@ -657,3 +711,7 @@ function getValidFeatureFlagSelectors(selectors?: SettingSelector[]): SettingSel
         return getValidSelectors(selectors);
     }
 }
+
+function isFailoverableError(error: any): boolean {
+    return isRestError(error) && (error.statusCode === 408 || error.statusCode === 429 || (error.statusCode !== undefined && error.statusCode >= 500));
```
zhiyuanliang-ms, line 716: https://github.com/Azure/AppConfiguration-DotnetProvider/issues/568

We treat 401 & 403 as failoverable error too.

src/AzureAppConfigurationImpl.ts
```diff
@@ -176,34 +176,70 @@ export class AzureAppConfigurationImpl implements AzureAppConfiguration {
...
+        throw new Error("All app configuration clients failed to get settings.");
+    }
 
-            for await (const setting of settings) {
-                if (!isFeatureFlag(setting)) { // exclude feature flags
-                    loadedSettings.push(setting);
+    async #loadSelectedKeyValues(): Promise<ConfigurationSetting[]> {
+        // validate selectors
+        const selectors = getValidKeyValueSelectors(this.#options?.selectors);
+        let loadedSettings: ConfigurationSetting[] = [];
```
zhiyuanliang-ms, line 217: Redundant allocation. Line 243 can just be 
``` javascript
return await this.#executeWithFailoverPolicy(funcToExecute);
```

src/AzureAppConfigurationImpl.ts
```diff
@@ -176,34 +176,70 @@ export class AzureAppConfigurationImpl implements AzureAppConfiguration {
...
+                };
+
+                const settings = listConfigurationSettingsWithTrace(
+                    this.#requestTraceOptions,
+                    client,
+                    listOptions
+                );
+
+                for await (const setting of settings) {
+                    if (!isFeatureFlag(setting)) { // exclude feature flags
```
zhiyuanliang-ms, line 234: Here is what I mentioned before that feature flags are loaded twice. The settings of feature flag have been loaded but they will be discarded. And you will call `loadFeatureFlag` again in the `load` method. @Eskibear 

But this is another issue which unrelated with this PR.

src/AzureAppConfigurationImpl.ts
zhiyuanliang-ms: Discussed offline. This is for deep copy. There is no elegant way to do deep copy in javascript.

src/AzureAppConfigurationImpl.ts
```diff
@@ -258,29 +294,43 @@ export class AzureAppConfigurationImpl implements AzureAppConfiguration {
...
             }
-            selector.pageEtags = pageEtags;
-        }
+
+            this.#featureFlagSelectors = selectors;
+            return featureFlagSettings;
+        };
+
+        let featureFlagSettings: ConfigurationSetting[] = [];
+        featureFlagSettings = await this.#executeWithFailoverPolicy(funcToExecute);
```
zhiyuanliang-ms, line 333: Why sperate this into two lines?
``` javascript
const featureFlagSettings = await this.#executeWithFailoverPolicy(funcToExecute);
```

src/AzureAppConfigurationImpl.ts
zhiyuanliang-ms, line 716: Why do you only check whether error.statusCode is undefined for 5xx?

src/AzureAppConfigurationImpl.ts
```diff
@@ -433,31 +483,31 @@ export class AzureAppConfigurationImpl implements AzureAppConfiguration {
...
-                this.#client,
-                listOptions
-            ).byPage();
-
-            for await (const page of pageIterator) {
-                if (page._response.status === 200) { // created or changed
-                    needRefresh = true;
-                    break;
+        const funcToExecute = async (client) => {
+            const needRefresh = false;
```
zhiyuanliang-ms, line 487: Why do you set needRefresh to be a const false? How could this pass the testcase? Could you check whether our testsuite missed something?

src/AzureAppConfigurationImpl.ts
```diff
@@ -433,31 +483,31 @@ export class AzureAppConfigurationImpl implements AzureAppConfiguration {
...
 
-            if (needRefresh) {
-                break; // short-circuit if result from any of the selectors is changed
-            }
-        }
-
+        const needRefresh: boolean = await this.#executeWithFailoverPolicy(funcToExecute);
         if (needRefresh) {
             try {
                 await this.#loadFeatureFlags();
```
zhiyuanliang-ms, line 513: loadFeatureFlags should also be guarded with failover policy.

src/AzureAppConfigurationImpl.ts
```diff
@@ -433,31 +483,31 @@ export class AzureAppConfigurationImpl implements AzureAppConfiguration {
...
-            ).byPage();
-
-            for await (const page of pageIterator) {
-                if (page._response.status === 200) { // created or changed
-                    needRefresh = true;
-                    break;
+        const funcToExecute = async (client) => {
+            const needRefresh = false;
+            for (const selector of this.#featureFlagSelectors) {
+                const listOptions: ListConfigurationSettingsOptions = {
```
zhiyuanliang-ms, line 489: Why do you specify type for `listOptions`? This is the only place you specify the type in the whole function. Just curious about it, why treat it so special? And I strongly recommend to specify type for all non-primitive types in all functions.

src/AzureAppConfigurationOptions.ts
```diff
@@ -12,6 +12,14 @@ export const MaxRetries = 2;
 export const MaxRetryDelayInMs = 60000;
 
 export interface AzureAppConfigurationOptions {
+    /**
+     * Specifies whether enable replica discovery or not.
+     *
+     * @remarks
+     * If not specified, the default value is true.
+     */
+    replicaDiscoveryEnabled?: boolean;
```
zhiyuanliang-ms, line 21: Could you move this to the end?

src/AzureAppConfigurationOptions.ts
```diff
@@ -12,6 +12,14 @@ export const MaxRetries = 2;
 export const MaxRetryDelayInMs = 60000;
 
 export interface AzureAppConfigurationOptions {
+    /**
+     * Specifies whether enable replica discovery or not.
```
zhiyuanliang-ms, line 16:
```suggestion
     * Specifies whether to enable replica discovery or not.

```

src/AzureAppConfigurationOptions.ts
```diff
@@ -12,6 +12,14 @@ export const MaxRetries = 2;
...
+    /**
+     * Specifies whether enable replica discovery or not.
+     *
+     * @remarks
+     * If not specified, the default value is true.
+     */
+    replicaDiscoveryEnabled?: boolean;
+
     /**
      * Specify what key-values to include in the configuration provider.
```
zhiyuanliang-ms, line 24:
```suggestion
     * Specifies what key-values to include in the configuration provider.

```

Could you help update this please?

src/ConfigurationClientManager.ts
```diff
@@ -0,0 +1,317 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT license.
+
+import { AppConfigurationClient, AppConfigurationClientOptions } from "@azure/app-configuration";
+import { ConfigurationClientWrapper } from "./ConfigurationClientWrapper";
+import { TokenCredential } from "@azure/identity";
+import { AzureAppConfigurationOptions, MaxRetries, MaxRetryDelayInMs } from "./AzureAppConfigurationOptions";
+import { isFailoverableEnv } from "./requestTracing/utils";
+import * as RequestTracing from "./requestTracing/constants";
```
zhiyuanliang-ms, line 9: If you use namespace import here, could you also update the `./requestTracing/utils.ts` file to make them consistent.

src/ConfigurationClientManager.ts
```diff
@@ -0,0 +1,317 @@
...
+const ALT = "_alt";
+const TCP = "_tcp";
+const Endpoint = "Endpoint";
+const Id = "Id";
+const Secret = "Secret";
+const AzConfigDomainLabel = ".azconfig.";
+const AppConfigDomainLabel = ".appconfig.";
+const FallbackClientRefreshExpireInterval = 60 * 60 * 1000; // 1 hour in milliseconds
+const MinimalClientRefreshInterval = 30 * 1000; // 30 seconds in milliseconds
+const SrvQueryTimeout = 5000; // 5 seconds
```
zhiyuanliang-ms, line 21:
```suggestion
const SrvQueryTimeout = 5 * 1000; // 5 seconds in milliseconds
```
Make it consistent please.

src/ConfigurationClientManager.ts
```diff
@@ -0,0 +1,317 @@
...
+
+interface IConfigurationClientManager {
```
zhiyuanliang-ms, line 23: What's the reason to have this interface? If you create this, you should use it as the type for `#clientManager` of `AzureAppConfigurationImpl`.

src/ConfigurationClientManager.ts
```diff
@@ -0,0 +1,317 @@
...
+        }
+    }
+
+    return results;
+}
+
+/**
+ * Parses the connection string to extract the value associated with a specific token.
+ */
+function parseConnectionString(connectionString, token: string): string {
```
zhiyuanliang-ms, line 229: This function is not very elegant. Could you refer [this](https://github.com/Azure/azure-sdk-for-js/blob/main/sdk/appconfiguration/app-configuration/src/appConfigurationClient.ts#L167)?

src/load.ts
```diff
@@ -31,43 +30,37 @@ export async function load(
...
                     throw new Error("Invalid endpoint URL.", { cause: error });
                 } else {
                     throw error;
                 }
             }
-        } else {
-            clientEndpoint = connectionStringOrEndpoint.toString();
         }
         const credential = credentialOrOptions as TokenCredential;
         options = appConfigOptions;
```
zhiyuanliang-ms, line 56: This line is meaningless.

src/ConfigurationClientWrapper.ts
```diff
@@ -0,0 +1,49 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT license.
+
+import { AppConfigurationClient } from "@azure/app-configuration";
+
+const MaxBackoffDuration = 10 * 60 * 1000; // 10 minutes in milliseconds
+const MinBackoffDuration = 30 * 1000; // 30 seconds in milliseconds
+const MAX_SAFE_EXPONENTIAL = 30; // Used to avoid overflow. bitwise operations in JavaScript are limited to 32 bits. It overflows at 2^31 - 1.
+const JITTER_RATIO = 0.25;
```
zhiyuanliang-ms, line 9: These are all defined in `refresh/RefreshTimer.ts`.

src/ConfigurationClientWrapper.ts
```diff
@@ -0,0 +1,49 @@
...
+    if (successfull) {
+        clientWrapper.failedAttempts = 0;
+        clientWrapper.backoffEndTime = Date.now();
+    } else {
+        clientWrapper.failedAttempts += 1;
+        clientWrapper.backoffEndTime = Date.now() + calculateBackoffDuration(clientWrapper.failedAttempts);
+    }
+}
+
+export function calculateBackoffDuration(failedAttempts: number) {
```
zhiyuanliang-ms, line 33: Can you unify this function with the `calculateBackOffTime` function defined in `refresh/RefreshTimer.ts`? Most of the code are just the same. Please do not write repeated code.

src/ConfigurationClientWrapper.ts
```diff
@@ -0,0 +1,49 @@
...
+
+const MaxBackoffDuration = 10 * 60 * 1000; // 10 minutes in milliseconds
+const MinBackoffDuration = 30 * 1000; // 30 seconds in milliseconds
+const MAX_SAFE_EXPONENTIAL = 30; // Used to avoid overflow. bitwise operations in JavaScript are limited to 32 bits. It overflows at 2^31 - 1.
+const JITTER_RATIO = 0.25;
+
+export class ConfigurationClientWrapper {
+    endpoint: string;
+    client: AppConfigurationClient;
+    backoffEndTime: number = 0; // Timestamp
```
zhiyuanliang-ms, line 14: We've already have `RefreshTimer` class. You can modify it to make it more general for all backoff scenarios. ClientWrapper could own a refreshTimer and there is no need to re-implement the exisiting thing.

src/requestTracing/constants.ts
```diff
@@ -43,6 +43,7 @@ export enum RequestType {
     STARTUP = "Startup",
     WATCH = "Watch"
 }
+export const FAILOVER_REQUEST_TYPE = "FailoverRequest";
```
zhiyuanliang-ms, line 46: The tag is wrong. Please check .net provider.

https://github.com/Azure/AppConfiguration-DotnetProvider/blob/preview/src/Microsoft.Extensions.Configuration.AzureAppConfiguration/Constants/RequestTracingConstants.cs#L35

src/ConfigurationClientManager.ts
```diff
@@ -0,0 +1,317 @@
...
+        }
+
+        const currentTime = Date.now();
+        if (this.#isFallbackClientDiscoveryDue(currentTime)) {
+            this.#lastFallbackClientRefreshAttempt = currentTime;
+            const host = new URL(this.endpoint).hostname;
+            await this.#discoverFallbackClients(host);
+        }
+
+        // Filter static clients where BackoffEndTime is less than or equal to now
```
zhiyuanliang-ms, line 87:
```suggestion
        // Filter static clients whose backoff time has ended
```

src/ConfigurationClientManager.ts
```diff
@@ -0,0 +1,317 @@
...
+
+/**
+ * Query SRV records and return target hosts.
+ */
+async function querySrvTargetHost(host: string): Promise<string[]> {
+    const results: string[] = [];
+    let dns;
+
+    if (isFailoverableEnv()) {
+        dns = require("dns/promises");
```
zhiyuanliang-ms, line 172: We are following ESM syntax. You should always use import by default.

Here, you can 

```
dns = global.dns;
```

src/ConfigurationClientManager.ts
zhiyuanliang-ms: > Does it mean each time of querying srv target host, it will attempt to load the module?

Yes. That's why people write `const dns = require("dns")` at the beginning of the file.

src/ConfigurationClientManager.ts
```diff
@@ -0,0 +1,317 @@
...
+}
+
+/**
+ * Query SRV records and return target hosts.
+ */
+async function querySrvTargetHost(host: string): Promise<string[]> {
+    const results: string[] = [];
+    let dns;
+
+    if (isFailoverableEnv()) {
```
zhiyuanliang-ms, line 171: Can you make this function private method of `ConfigurationClientManager` so that you won't need to do the redundant check for failoverable env. According to the path, this function is only called in `#discoverFallbackClient` which will only be called when this.isFailoverable is true.

src/AzureAppConfigurationImpl.ts
linglingye001: `isFailoverRequest` is updated after each request. Once the current client succeeded, it will be set to `false` and next time the request will not be tagged failover.

src/AzureAppConfigurationImpl.ts
linglingye001: Two `refreshclients` calls for different situation. Since client has backOffTime so we may get empty client list when call `getClients`. `refreshClients` in line 187 is to discover new replicas when no available clients to send any request. The other refresh call in line 211 is also for new replica discovery but happens after all the available clients failed.

src/AzureAppConfigurationImpl.ts
```diff
@@ -176,34 +176,70 @@ export class AzureAppConfigurationImpl implements AzureAppConfiguration {
...
+        this.#clientManager.refreshClients();
+        throw new Error("All app configuration clients failed to get settings.");
+    }
 
-            for await (const setting of settings) {
-                if (!isFeatureFlag(setting)) { // exclude feature flags
-                    loadedSettings.push(setting);
+    async #loadSelectedKeyValues(): Promise<ConfigurationSetting[]> {
+        // validate selectors
+        const selectors = getValidKeyValueSelectors(this.#options?.selectors);
```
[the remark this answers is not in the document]
linglingye001: funcToExecute may be executed multiple times when iterating through the clients, putting selectors outside can avoid repeated calls when failover happens.

src/AzureAppConfigurationImpl.ts
linglingye001: `===` operation will check if statusCode is undefined but `>=` not.

zhiyuanliang-ms: image

zhiyuanliang-ms: Why not just error.statusCode >= 500?

src/AzureAppConfigurationImpl.ts
zhiyuanliang-ms: Then, why do we need to `refreshClients` in line 210? Next time calling #executeWithFailoverPolicy will call `getClients` where it will also try to discover fallback clients?

src/ConfigurationClientManager.ts
```diff
@@ -0,0 +1,317 @@
...
+                    .filter(client => client.backoffEndTime <= currentTime));
+        }
+
+        return availableClients;
+    }
+
+    async refreshClients() {
+        const currentTime = Date.now();
+        if (this.isFailoverable &&
+            currentTime > new Date(this.#lastFallbackClientRefreshAttempt + MinimalClientRefreshInterval).getTime()) {
```
zhiyuanliang-ms, line 102: Why don't you use `this.#isFallbackClientDiscoveryDue` here like what you did in `getClients`?

src/ConfigurationClientManager.ts
```diff
@@ -0,0 +1,317 @@
...
+            this.#lastFallbackClientRefreshAttempt = currentTime;
+            const url = new URL(this.endpoint);
```
zhiyuanliang-ms, line 104: Can you make it consistent with what you did in `getClients`:

``` js
const host = new URL(this.endpoint).hostname;
await this.#discoverFallbackClients(host);
```

src/ConfigurationClientManager.ts
linglingye001: refreshClients will be called if no availabel clients or all clients failed, these scenarios can happen when fallback isn't due.

package-lock.json
linglingye001: maybe

src/AzureAppConfigurationImpl.ts
linglingye001: discussed offline

src/ConfigurationClientManager.ts
```diff
@@ -0,0 +1,310 @@
...
+import * as RequestTracing from "./requestTracing/constants.js";
+
+const TCP_ORIGIN_KEY_NAME = "_origin._tcp";
+const ALT_KEY_NAME = "_alt";
+const TCP_KEY_NAME = "_tcp";
+const ENDPOINT_KEY_NAME = "Endpoint";
+const ID_KEY_NAME = "Id";
+const SECRET_KEY_NAME = "Secret";
+const AZCONFIG_DOMAIN_LABEL = ".azconfig.";
+const APPCONFIG_DOMAIN_LABEL = ".appconfig.";
```
zhiyuanliang-ms, line 18:
```suggestion
const TRUSTED_DOMAIN_LABELS = [".azconfig.", ".appconfig."];
```

Then you won't need line 257

src/ConfigurationClientManager.ts
```diff
@@ -0,0 +1,310 @@
...
+            }
+
+            this.#dynamicClients = newDynamicClients;
+            this.#lastFallbackClientRefreshTime = Date.now();
+        } catch (err) {
+            console.warn(`Fail to build fallback clients, ${err.message}`);
+        }
+    }
+
+    #newConfigurationClient(endpoint) {
```
zhiyuanliang-ms, line 160: This method is only used once. And the method name is confusing. No need to create such a method.  Please write it directly at line 148.

src/ConfigurationClientManager.ts
```diff
@@ -0,0 +1,310 @@
...
+                        continue;
+                    }
+                    const client = this.#newConfigurationClient(targetEndpoint);
+                    newDynamicClients.push(new ConfigurationClientWrapper(targetEndpoint, client));
+                }
+            }
+
+            this.#dynamicClients = newDynamicClients;
+            this.#lastFallbackClientRefreshTime = Date.now();
+        } catch (err) {
```
zhiyuanliang-ms, line 155: We should only allow querySrv to fail. Most of the code in try {} should be put outside.

src/ConfigurationClientManager.ts
```diff
@@ -0,0 +1,310 @@
...
+        }
+    }
+
+    async #discoverFallbackClients(host) {
+        const timeout = setTimeout(() => {
+        }, SRV_QUERY_TIMEOUT);
+        const srvResults = await querySrvTargetHost(host);
+
+        try {
+            const result = await Promise.race([srvResults, timeout]);
```
zhiyuanliang-ms, line 128:
``` js
const srvResults = await querySrvTargetHost(host);
try {
    const result = await Promise.race([srvResults, timeout]);
```

Are you sure this is correct? Promise.race should take `Promise` as parameter.

zhiyuanliang-ms (changes requested):
const srvResults = await querySrvTargetHost(host);
try {
    const result = await Promise.race([srvResults, timeout]);
Are you sure this is correct? Promise.race should take Promise as parameter.

src/ConfigurationClientManager.ts
```diff
@@ -0,0 +1,310 @@
...
+const ALT_KEY_NAME = "_alt";
+const TCP_KEY_NAME = "_tcp";
+const ENDPOINT_KEY_NAME = "Endpoint";
+const ID_KEY_NAME = "Id";
+const SECRET_KEY_NAME = "Secret";
+const AZCONFIG_DOMAIN_LABEL = ".azconfig.";
+const APPCONFIG_DOMAIN_LABEL = ".appconfig.";
+const FALLBACK_CLIENT_REFRESH_EXPIRE_INTERVAL = 60 * 60 * 1000; // 1 hour in milliseconds
+const MINIMAL_CLIENT_REFRESH_INTERVAL = 30 * 1000; // 30 seconds in milliseconds
+const SRV_QUERY_TIMEOUT = 5* 1000; // 5 seconds in milliseconds
```
zhiyuanliang-ms, line 21:
```suggestion
const SRV_QUERY_TIMEOUT = 5 * 1000; // 5 seconds in milliseconds
```

zhiyuanliang-ms, line 21: .NET provider used 30s timeout. [ref](https://github.com/Azure/AppConfiguration-DotnetProvider/blob/main/src/Microsoft.Extensions.Configuration.AzureAppConfiguration/ConfigurationClientManager.cs#L192)

@RichardChen820 Can you confirm the behavior of k8s provider?

src/ConfigurationClientManager.ts
```diff
@@ -0,0 +1,310 @@
...
+        }
+
+        const connectionStr = buildConnectionString(endpoint, this.#secret, this.#id);
+        return new AppConfigurationClient(connectionStr, this.#clientOptions);
+    }
+
+    #isFallbackClientDiscoveryDue(dateTime) {
+        return dateTime >= this.#lastFallbackClientRefreshAttempt + MINIMAL_CLIENT_REFRESH_INTERVAL
+            && (!this.#dynamicClients
+                || this.#dynamicClients.every(client => dateTime < client.backoffEndTime)
```
zhiyuanliang-ms, line 172: Should it be > client.backoffEndTime?

src/ConfigurationClientManager.ts
```diff
@@ -0,0 +1,310 @@
...
+        // Add the first origin record to results
+        const originHost = originRecords[0].name;
+        results.push(originHost);
+
+        // Look up SRV records for alternate hosts
+        let index = 0;
+        let moreAltRecordsExist = true;
+        while (moreAltRecordsExist) {
+            const currentAlt = `${ALT_KEY_NAME}${index}`;
+            try {
```
zhiyuanliang-ms, line 206: Why do you have nested try-catch in this function?

src/ConfigurationClientManager.ts
```diff
@@ -0,0 +1,310 @@
...
+        }
+
+        // Add the first origin record to results
+        const originHost = originRecords[0].name;
+        results.push(originHost);
+
+        // Look up SRV records for alternate hosts
+        let index = 0;
+        let moreAltRecordsExist = true;
+        while (moreAltRecordsExist) {
```
zhiyuanliang-ms, line 204:
```suggestion
        // eslint-disable-next-line no-constant-condition
        while (true) {
```

And remove `moreAltRecordsExist`

src/ConfigurationClientManager.ts
```diff
@@ -0,0 +1,310 @@
...
+
+    if (typeof global !== "undefined" && global.dns) {
+        dns = global.dns;
+    } else {
+        throw new Error("Failover is not supported in the current environment.");
+    }
+
+    try {
+        // Look up SRV records for the origin host
+        const originRecords = await dns.resolveSrv(`${TCP_ORIGIN_KEY_NAME}.${host}`);
```
zhiyuanliang-ms, line 192: Can you double check whether the current usage of dns.resolveSrv is correct? It doesn't match the [doc](https://nodejs.org/api/dns.html#dnsresolvesrvhostname-callback)

I tested it and got the error:
image

src/ConfigurationClientManager.ts
```diff
@@ -0,0 +1,310 @@
...
+
+/**
+ * Query SRV records and return target hosts.
+ */
+async function querySrvTargetHost(host: string): Promise<string[]> {
+    const results: string[] = [];
+    let dns;
+
+    if (typeof global !== "undefined" && global.dns) {
+        dns = global.dns;
```
zhiyuanliang-ms, line 185: I just found that we cannot use global.dns in script enviroment. It only works in Node.js shell environment.
We have to write something like:
``` js
const dns = await import("dns");
```

And in this case, it should be a part of the process determining whether current environment supports replica auto discovery. We should only import for once and the loaded module should be stored as a property of the client manager.

src/requestTracing/utils.ts
```diff
@@ -165,3 +172,11 @@ function isWebWorker() {
 
     return workerGlobalScopeDefined && importScriptsAsGlobalFunction && isNavigatorDefinedAsExpected;
 }
+
+export function isFailoverableEnv() {
```
zhiyuanliang-ms, line 176: This function should not be put under the requestTracing folder.

zhiyuanliang-ms (changes requested):
The replica discovery is completely dependent on `dns` module. So whether `dns` module can be imported should be the criteria whether current environment supports replica discovery. 
Ideally, the we should decide whether current environment supports failover in the constructor of client manager. However, we have to use async call `await import("dns")` to decide whether `dns` module can be loaded. So it cannot be put in the constructor.

Suggestion:
Add an async `init` method for `ClientManager`. After you create `const xxx = new ClientManager()`, call `await xxx.init()`. 
In the `init` method, it should determine whether `dns` module can be loaded, i.e. whether replica discovery is available.

Note:
1. await import("xxx") will throw exception if module cannot be found. Need to handle the error. If `dns` cannot be loaded, then client manager should disable failover.
2. the `dns` module should only be loaded once.

src/requestTracing/constants.ts
```diff
@@ -43,6 +43,7 @@ export enum RequestType {
     STARTUP = "Startup",
     WATCH = "Watch"
 }
+export const FAILOVER_REQUEST_TAG = "Failover";
```
zhiyuanliang-ms, line 46: Could you move this line after line 49, since we have a comment line `//Tag names` in line 49

src/requestTracing/utils.ts
```diff
@@ -100,6 +103,10 @@ export function createCorrelationContextHeader(options: AzureAppConfigurationOpt
         contextParts.push(tag);
     }
 
+    if (isFailoverRequest) {
```
zhiyuanliang-ms, line 106: Did you add any testcase for this?

src/ConfigurationClientWrapper.ts
```diff
@@ -0,0 +1,49 @@
...
+        clientWrapper.backoffEndTime = Date.now() + calculateBackoffDuration(clientWrapper.failedAttempts);
+    }
+}
+
+export function calculateBackoffDuration(failedAttempts: number) {
+    if (failedAttempts <= 1) {
+        return MinBackoffDuration;
+    }
+
+    // exponential: minBackoff * 2^(failedAttempts-1)
```
zhiyuanliang-ms, line 38:
```suggestion
    // exponential: minBackoff * 2 ^ (failedAttempts - 1)
```

src/ConfigurationClientWrapper.ts
```diff
@@ -0,0 +1,49 @@
...
+    backoffEndTime: number = 0; // Timestamp
+    failedAttempts: number = 0;
+
+    constructor(endpoint: string, client: AppConfigurationClient) {
+        this.endpoint = endpoint;
+        this.client = client;
+    }
+}
+
+export function updateClientBackoffStatus(clientWrapper: ConfigurationClientWrapper, successfull: boolean) {
```
zhiyuanliang-ms, line 23: It is more appropriate to let this function be a public method of `ConfigurationClientWrapper` class.

src/ConfigurationClientManager.ts
```diff
@@ -0,0 +1,296 @@
...
+const SECRET_KEY_NAME = "Secret";
+const TRUSTED_DOMAIN_LABELS = [".azconfig.", ".appconfig."];
+const FALLBACK_CLIENT_REFRESH_EXPIRE_INTERVAL = 60 * 60 * 1000; // 1 hour in milliseconds
+const MINIMAL_CLIENT_REFRESH_INTERVAL = 30 * 1000; // 30 seconds in milliseconds
+const SRV_QUERY_TIMEOUT = 30 * 1000; // 30 seconds in milliseconds
+
+export class ConfigurationClientManager {
+    isFailoverable: boolean;
+    dns: any;
+    endpoint: string;
```
zhiyuanliang-ms, line 25: It seems like there is no where outside of this class using these properties. Then we should set them private.

src/ConfigurationClientManager.ts
```diff
@@ -0,0 +1,296 @@
...
+            const ConnectionStringRegex = /Endpoint=(.*);Id=(.*);Secret=(.*)/;
+            const regexMatch = connectionString.match(ConnectionStringRegex);
+            if (regexMatch) {
+                this.endpoint = regexMatch[1];
+                this.#id = regexMatch[2];
+                this.#secret = regexMatch[3];
+            } else {
+                throw new Error(`Invalid connection string. Valid connection strings should match the regex '${ConnectionStringRegex.source}'.`);
+            }
+        } else if ((connectionStringOrEndpoint instanceof URL || typeof connectionStringOrEndpoint === "string") && instanceOfTokenCredential(credentialOrOptions)) {
```
zhiyuanliang-ms, line 58: We can use a variable to store the result of `instanceOfTokenCredential(credentialOrOptions)`

In this case, we can avoid the duplicated call of this method if the code path falls into the else if branch.

src/ConfigurationClientManager.ts
```diff
@@ -0,0 +1,296 @@
...
+
+    async #discoverFallbackClients(host) {
+        let result;
+        try {
+            result = await Promise.race([
+                new Promise((_, reject) => setTimeout(() => reject(new Error("SRV record query timed out.")), SRV_QUERY_TIMEOUT)),
+                this.#querySrvTargetHost(host)
+            ]);
+        } catch (error) {
+            throw new Error(`Fail to build fallback clients, ${error.message}`);
```
zhiyuanliang-ms, line 146:
```suggestion
            throw new Error(`Failed to build fallback clients, ${error.message}`);
```

src/ConfigurationClientManager.ts
```diff
@@ -0,0 +1,296 @@
...
+                        throw error;
+                    }
+                }
+            }
+
+            const credential = credentialOrOptions as TokenCredential;
+            this.#appConfigOptions = appConfigOptions as AzureAppConfigurationOptions;
+            this.#clientOptions = getClientOptions(this.#appConfigOptions);
+            staticClient = new AppConfigurationClient(connectionStringOrEndpoint.toString(), credential, this.#clientOptions);
+            this.endpoint = endpoint.toString();
```
zhiyuanliang-ms, line 77: This line is no longer needed. The type of `endpoint` property should be `URL`. In this case, you can get rid of the try-catch in the `getValidDomain` function.

src/ConfigurationClientManager.ts
```diff
@@ -0,0 +1,296 @@
...
+const ID_KEY_NAME = "Id";
+const SECRET_KEY_NAME = "Secret";
+const TRUSTED_DOMAIN_LABELS = [".azconfig.", ".appconfig."];
+const FALLBACK_CLIENT_REFRESH_EXPIRE_INTERVAL = 60 * 60 * 1000; // 1 hour in milliseconds
+const MINIMAL_CLIENT_REFRESH_INTERVAL = 30 * 1000; // 30 seconds in milliseconds
+const SRV_QUERY_TIMEOUT = 30 * 1000; // 30 seconds in milliseconds
+
+export class ConfigurationClientManager {
+    isFailoverable: boolean;
+    dns: any;
```
zhiyuanliang-ms, line 24: It seems like there is no where outside of this class using these properties. Then we should set them private.

src/ConfigurationClientManager.ts
zhiyuanliang-ms, line 25: Suggest to use type URL here.

src/ConfigurationClientWrapper.ts
```diff
@@ -0,0 +1,49 @@
...
+
+import { AppConfigurationClient } from "@azure/app-configuration";
+
+const MaxBackoffDuration = 10 * 60 * 1000; // 10 minutes in milliseconds
+const MinBackoffDuration = 30 * 1000; // 30 seconds in milliseconds
+const MAX_SAFE_EXPONENTIAL = 30; // Used to avoid overflow. bitwise operations in JavaScript are limited to 32 bits. It overflows at 2^31 - 1.
+const JITTER_RATIO = 0.25;
+
+export class ConfigurationClientWrapper {
+    endpoint: string;
```
zhiyuanliang-ms, line 12: Is this used in any place? Suggest removing it.

src/ConfigurationClientWrapper.ts
```diff
@@ -0,0 +1,49 @@
...
+    client: AppConfigurationClient;
+    backoffEndTime: number = 0; // Timestamp
+    failedAttempts: number = 0;
```
zhiyuanliang-ms, line 15: Should be private?

src/ConfigurationClientManager.ts
```diff
@@ -0,0 +1,296 @@
...
+            const regexMatch = connectionString.match(ConnectionStringRegex);
+            if (regexMatch) {
+                this.endpoint = regexMatch[1];
+                this.#id = regexMatch[2];
+                this.#secret = regexMatch[3];
+            } else {
+                throw new Error(`Invalid connection string. Valid connection strings should match the regex '${ConnectionStringRegex.source}'.`);
+            }
+        } else if ((connectionStringOrEndpoint instanceof URL || typeof connectionStringOrEndpoint === "string") && instanceOfTokenCredential(credentialOrOptions)) {
+            let endpoint = connectionStringOrEndpoint;
```
zhiyuanliang-ms, line 59: This line is no longer needed.

src/ConfigurationClientManager.ts
```diff
@@ -0,0 +1,296 @@
...
+        credentialOrOptions?: TokenCredential | AzureAppConfigurationOptions,
+        appConfigOptions?: AzureAppConfigurationOptions
+    ) {
+        let staticClient: AppConfigurationClient;
+
+        if (typeof connectionStringOrEndpoint === "string" && !instanceOfTokenCredential(credentialOrOptions)) {
+            const connectionString = connectionStringOrEndpoint;
+            this.#appConfigOptions = credentialOrOptions as AzureAppConfigurationOptions;
+            this.#clientOptions = getClientOptions(this.#appConfigOptions);
+            staticClient = new AppConfigurationClient(connectionString, this.#clientOptions);
```
zhiyuanliang-ms, line 48: This line should be placed after the error throwing of invalid connection string.

src/ConfigurationClientManager.ts
```diff
@@ -0,0 +1,296 @@
...
+                    } else {
+                        throw error;
+                    }
+                }
+            }
+
+            const credential = credentialOrOptions as TokenCredential;
+            this.#appConfigOptions = appConfigOptions as AzureAppConfigurationOptions;
+            this.#clientOptions = getClientOptions(this.#appConfigOptions);
+            staticClient = new AppConfigurationClient(connectionStringOrEndpoint.toString(), credential, this.#clientOptions);
```
zhiyuanliang-ms, line 76: Move it to the bottom of this else-if branch.
And 
```suggestion
            staticClient = new AppConfigurationClient(this.endpoint.toString(), this.#credential, this.#clientOptions);
```

src/ConfigurationClientManager.ts
```diff
@@ -0,0 +1,296 @@
...
+            result = await Promise.race([
+                new Promise((_, reject) => setTimeout(() => reject(new Error("SRV record query timed out.")), SRV_QUERY_TIMEOUT)),
+                this.#querySrvTargetHost(host)
+            ]);
+        } catch (error) {
+            throw new Error(`Fail to build fallback clients, ${error.message}`);
+        }
+
+        const srvTargetHosts = result as string[];
+        // Shuffle the list of SRV target hosts
```
zhiyuanliang-ms, line 150: Different from the .net provider's [implementation](https://github.com/Azure/AppConfiguration-DotnetProvider/blob/main/src/Microsoft.Extensions.Configuration.AzureAppConfiguration/ConfigurationClientManager.cs#L239), missed the sorting based on weight and priority.

@RichardChen820 Can you help double check this?

src/ConfigurationClientManager.ts
zhiyuanliang-ms: .NET provider used `<=` [ref](https://github.com/Azure/AppConfiguration-DotnetProvider/blob/main/src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationProvider.cs#L224)

src/ConfigurationClientManager.ts
```diff
@@ -0,0 +1,296 @@
...
+        }
+
+        this.#dynamicClients = newDynamicClients;
+        this.#lastFallbackClientRefreshTime = Date.now();
+    }
+
+    #isFallbackClientDiscoveryDue(dateTime) {
+        return dateTime >= this.#lastFallbackClientRefreshAttempt + MINIMAL_CLIENT_REFRESH_INTERVAL
+            && (!this.#dynamicClients
+                || this.#dynamicClients.every(client => dateTime < client.backoffEndTime)
```
zhiyuanliang-ms, line 175: Seems to be different from .NET provider's implementation [ref](https://github.com/Azure/AppConfiguration-DotnetProvider/blob/main/src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationProvider.cs#L224)

I forget the reason you told me why here is `<`. Could you explain it here in case I forget it again?

src/ConfigurationClientManager.ts
```diff
@@ -0,0 +1,296 @@
...
+                const client = this.#credential ? new AppConfigurationClient(targetEndpoint, this.#credential, this.#clientOptions) : new AppConfigurationClient(buildConnectionString(targetEndpoint, this.#secret, this.#id), this.#clientOptions);
+                newDynamicClients.push(new ConfigurationClientWrapper(targetEndpoint, client));
+            }
+        }
+
+        this.#dynamicClients = newDynamicClients;
+        this.#lastFallbackClientRefreshTime = Date.now();
+    }
+
+    #isFallbackClientDiscoveryDue(dateTime) {
```
zhiyuanliang-ms, line 172: Suggest not passing dataTime as parameter and useing `Date.now()` directly (IMO, more readable)

src/ConfigurationClientManager.ts
```diff
@@ -0,0 +1,296 @@
...
+
+        return availableClients;
+    }
+
+    async refreshClients() {
+        const currentTime = Date.now();
+        if (this.isFailoverable &&
+            currentTime > new Date(this.#lastFallbackClientRefreshAttempt + MINIMAL_CLIENT_REFRESH_INTERVAL).getTime()) {
+            this.#lastFallbackClientRefreshAttempt = currentTime;
+            const host = new URL(this.endpoint).hostname;
```
zhiyuanliang-ms, line 133: If `this.endpoint` is of type `URL`, then this new URL() can be saved.

src/ConfigurationClientManager.ts
```diff
@@ -0,0 +1,296 @@
...
+                    .filter(client => client.backoffEndTime <= currentTime));
+        }
+
+        return availableClients;
+    }
+
+    async refreshClients() {
+        const currentTime = Date.now();
+        if (this.isFailoverable &&
+            currentTime > new Date(this.#lastFallbackClientRefreshAttempt + MINIMAL_CLIENT_REFRESH_INTERVAL).getTime()) {
```
zhiyuanliang-ms, line 131: Why did you use a different implementation from what you wrote in `isFallbackClientDiscoveryDue` function?
```
dateTime >= this.#lastFallbackClientRefreshAttempt + MINIMAL_CLIENT_REFRESH_INTERVAL
```

src/ConfigurationClientManager.ts
```diff
@@ -0,0 +1,296 @@
...
+                new Promise((_, reject) => setTimeout(() => reject(new Error("SRV record query timed out.")), SRV_QUERY_TIMEOUT)),
+                this.#querySrvTargetHost(host)
+            ]);
+        } catch (error) {
+            throw new Error(`Fail to build fallback clients, ${error.message}`);
+        }
+
+        const srvTargetHosts = result as string[];
+        // Shuffle the list of SRV target hosts
+        for (let i = srvTargetHosts.length - 1; i > 0; i--) {
```
zhiyuanliang-ms, line 151: Suggest making this shuffle implementation a helper method and placing it in a utils file.

src/ConfigurationClientManager.ts
```diff
@@ -0,0 +1,296 @@
...
+        }
+
+        const newDynamicClients: ConfigurationClientWrapper[] = [];
+        for (const host of srvTargetHosts) {
+            if (isValidEndpoint(host, this.#validDomain)) {
+                const targetEndpoint = `https://${host}`;
+                if (targetEndpoint.toLowerCase() === this.endpoint.toLowerCase()) {
+                    continue;
+                }
+                const client = this.#credential ? new AppConfigurationClient(targetEndpoint, this.#credential, this.#clientOptions) : new AppConfigurationClient(buildConnectionString(targetEndpoint, this.#secret, this.#id), this.#clientOptions);
```
zhiyuanliang-ms, line 163:
```suggestion
                const client = this.#credential ? 
                               new AppConfigurationClient(targetEndpoint, this.#credential, this.#clientOptions) : 
                               new AppConfigurationClient(buildConnectionString(targetEndpoint, this.#secret, this.#id), this.#clientOptions);
```

zhiyuanliang-ms, line 163: Can you add line break after "?" and ":" since this line is quite long.

src/ConfigurationClientManager.ts
zhiyuanliang-ms: I also didn't see this in .NET provider [ref](https://github.com/Azure/AppConfiguration-DotnetProvider/blob/main/src/Microsoft.Extensions.Configuration.AzureAppConfiguration/ConfigurationClientManager.cs#L99)

src/ConfigurationClientManager.ts
```diff
@@ -0,0 +1,296 @@
...
+            await this.#discoverFallbackClients(host);
+        }
+
+        // Filter static clients whose backoff time has ended
+        let availableClients = this.#staticClients.filter(client => client.backoffEndTime <= currentTime);
+        // If there are dynamic clients, filter and concatenate them
+        if (this.#dynamicClients && this.#dynamicClients.length > 0) {
+            availableClients = availableClients.concat(
+                this.#dynamicClients
+                    .filter(client => client.backoffEndTime <= currentTime));
```
zhiyuanliang-ms, line 122: For the code path where `isFallbackClientDiscoveryDue` is true and`discoverFallbackClients` is called, `this.#dynamicClients` will be all brand-new clients.

The `filter` call can be saved.

``` javascript
 
let availableClients = this.#staticClients.filter(client => client.backoffEndTime <= currentTime);
if (this.#isFallbackClientDiscoveryDue()){
    await this.#discoverFallbackClients();
    return availableClients.concat(this.#dynamicClients)
}
return availableClients.concat(this.#dynamicClients.filter(client => client.backoffEndTime <= currentTime));
```

src/AzureAppConfigurationImpl.ts
```diff
@@ -176,35 +176,70 @@ export class AzureAppConfigurationImpl implements AzureAppConfiguration {
...
         };
     }
 
-    async #loadSelectedKeyValues(): Promise<ConfigurationSetting[]> {
-        const loadedSettings: ConfigurationSetting[] = [];
+    async #executeWithFailoverPolicy(funcToExecute: (client: AppConfigurationClient) => Promise<any>): Promise<any> {
+        const clientWrappers = await this.#clientManager.getClients();
+        if (clientWrappers.length === 0) {
+            this.#clientManager.refreshClients();
+            throw new Error("No client is available to connect to the target app configuration store.");
```
zhiyuanliang-ms, line 188: This behavior is different from .NET provider [ref](https://github.com/Azure/AppConfiguration-DotnetProvider/blob/main/src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationProvider.cs#L232)

src/AzureAppConfigurationImpl.ts
```diff
@@ -176,35 +176,70 @@ export class AzureAppConfigurationImpl implements AzureAppConfiguration {
...
+            }
+        }
 
-            const settings = listConfigurationSettingsWithTrace(
-                this.#requestTraceOptions,
-                this.#client,
-                listOptions
-            );
+        this.#clientManager.refreshClients();
+        throw new Error("All app configuration clients failed to get settings.");
```
zhiyuanliang-ms, line 212:
```suggestion
        throw new Error("Failed to get settings from all App Configuration clients.");

```

src/AzureAppConfigurationImpl.ts
```diff
@@ -657,3 +696,8 @@ function getValidFeatureFlagSelectors(selectors?: SettingSelector[]): SettingSel
         return getValidSelectors(selectors);
     }
 }
+
+function isFailoverableError(error: any): boolean {
+    return isRestError(error) && (error.code === "ENOTFOUND" ||
```
zhiyuanliang-ms, line 701: Did you figure out the "ENOENT" error you met before?

test/featureFlag.test.ts
```diff
@@ -61,7 +61,9 @@ const mockedKVs = [{
 describe("feature flags", function () {
     this.timeout(10000);
 
+    const isFailoverable = false;
     before(() => {
+        mockConfigurationManagerGetClients(isFailoverable);
```
zhiyuanliang-ms, line 66: Do we really need this change?

src/ConfigurationClientWrapper.ts
linglingye001: For future load balance support

src/ConfigurationClientManager.ts
linglingye001: discussed offline

zhiyuanliang-ms: Note: if there is any client with backoffEndTime <= Date.now(0 which means there is available client, then no need to discover more

src/ConfigurationClientManager.ts
zhiyuanliang-ms: Related to this [conversation](https://github.com/Azure/AppConfiguration-JavaScriptProvider/pull/98/files#r1838657409)
This can be improved even more if you implement `isFallbackClientDiscoveryDue` inline here. 
Then, we can only do something like this:

``` javascript
const availableReplicas = this.#dynamicClients.filter(client => client.backoffEndTime <= currentTime);
if (dateTime >= this.#lastFallbackClientRefreshAttempt + MINIMAL_CLIENT_REFRESH_INTERVAL
            && (availableReplicas.length == 0 || Date.now() >= this.#lastFallbackClientRefreshTime + FALLBACK_CLIENT_REFRESH_EXPIRE_INTERVAL)) {
    await this.#discoverFallbackClients();
    return availableClients.concat(this.#dynamicClients)
}
return availableClients.concat(availableReplicas);
```

src/AzureAppConfigurationImpl.ts
linglingye001: This error is thrown when dns file modified, I think it's not typical for failover scenario. If customer encounter it in the future, we can update the failoverable error then.

test/featureFlag.test.ts
linglingye001: removed.

src/AzureAppConfigurationImpl.ts
linglingye001: updated

src/ConfigurationClientManager.ts
```diff
@@ -0,0 +1,296 @@
...
+        return "";
+    }
+
+    return `${ENDPOINT_KEY_NAME}=${endpoint};${ID_KEY_NAME}=${id};${SECRET_KEY_NAME}=${secret}`;
+}
+
+/**
+ * Extracts a valid domain from the given endpoint URL based on trusted domain labels.
+ */
+export function getValidDomain(endpoint: string): string {
```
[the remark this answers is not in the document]
linglingye001: for test case

src/AzureAppConfigurationImpl.ts
linglingye001: Added 'ENOENT' to failvoerable error.

src/AzureAppConfigurationImpl.ts
```diff
@@ -176,35 +175,70 @@ export class AzureAppConfigurationImpl implements AzureAppConfiguration {
...
         };
     }
 
-    async #loadSelectedKeyValues(): Promise<ConfigurationSetting[]> {
-        const loadedSettings: ConfigurationSetting[] = [];
+    async #executeWithFailoverPolicy(funcToExecute: (client: AppConfigurationClient) => Promise<any>): Promise<any> {
+        const clientWrappers = await this.#clientManager.getClients();
+        if (clientWrappers.length === 0) {
+            this.#clientManager.refreshClients();
+            console.warn("Refresh skipped because no endpoint is accessible.");
```
zhiyuanliang-ms, line 187: This is from .NET provider. But it doesn't make any sense. We actually do the refresh because no endpoint is accessible.

zhiyuanliang-ms: https://github.com/Azure/AppConfiguration-DotnetProvider/blob/main/src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationProvider.cs#L232

src/AzureAppConfigurationImpl.ts
```diff
@@ -176,35 +175,70 @@ export class AzureAppConfigurationImpl implements AzureAppConfiguration {
...
+        }
 
-        // validate selectors
-        const selectors = getValidKeyValueSelectors(this.#options?.selectors);
+        let successful: boolean;
+        for (const clientWrapper of clientWrappers) {
+            successful = false;
```
zhiyuanliang-ms, line 192: This line is wrong. No request will be marked as failover request then.

src/AzureAppConfigurationImpl.ts
zhiyuanliang-ms: In the .NET provider, the refresh means kv refresh operation. The .net implementation is different from here. The logic of getClients is not included in the executeWithFailoverPolicy. The message is specific for the `Refresh` method of provider.

zhiyuanliang-ms: If we we want to have the same behavior as .NET, we should return here and do nothing. Here the load and refresh scenarios are different. We can tolerant refresh doing nothing if there is no client available. No exception should be thrown, just log warning is fine. The load scenario is more important, we should make sure that configuration is pulled and if fail, throw exception.

test/refresh.test.ts
```diff
@@ -42,7 +42,8 @@ describe("dynamic refresh", function () {
 
     it("should throw error when refresh is not enabled but refresh is called", async () => {
         const connectionString = createMockedConnectionString();
-        const settings = await load(connectionString);
+        const settings = await load(connectionString, {
+        });
```
zhiyuanliang-ms, line 46: Why make this change?

test/refresh.test.ts
```diff
@@ -124,7 +125,8 @@ describe("dynamic refresh", function () {
 
     it("should throw error when calling onRefresh when refresh is not enabled", async () => {
         const connectionString = createMockedConnectionString();
-        const settings = await load(connectionString);
+        const settings = await load(connectionString, {
```
zhiyuanliang-ms, line 128: Same question

test/keyvault.test.ts
```diff
@@ -39,7 +39,8 @@ describe("key vault reference", function () {
     });
 
     it("require key vault options to resolve reference", async () => {
-        return expect(load(createMockedConnectionString())).eventually.rejectedWith("Configure keyVaultOptions to resolve Key Vault Reference(s).");
+        return expect(load(createMockedConnectionString(), {
```
zhiyuanliang-ms, line 42: Any reason to make this change?

test/utils/testHelper.ts
```diff
@@ -109,6 +131,57 @@ function mockAppConfigurationClientGetConfigurationSetting(kvList) {
     });
 }
 
+function mockAppConfigurationClientListConfigurationSettingsWithFailure(...pages: ConfigurationSetting[][]) {
+    const stub = sinon.stub(AppConfigurationClient.prototype, "listConfigurationSettings");
+
+    // Configure the stub to throw an error on the first call and return mockedKVs on the second call
+    stub.onFirstCall().throws(new RestError("Internal Server Error", { statusCode: 500 }));
+    stub.callsFake((listOptions) => {
```
zhiyuanliang-ms, line 139: These code is duplicated to line 55.

src/AzureAppConfigurationImpl.ts
linglingye001: discussed offline, log warning in refresh()

test/failover.test.ts
```diff
@@ -0,0 +1,122 @@
...
+    afterEach(() => {
+        restoreMocks();
+    });
+
+    it("should failover to replica and load key values from config store", async () => {
+        const replicaDiscoveryEnabled = true;
+        const isFailoverable = true;
+        mockConfigurationManagerGetClients(isFailoverable);
+        mockAppConfigurationClientListConfigurationSettingsWithFailure();
+        restoreMocks();
```
zhiyuanliang-ms, line 41: This is wrong. It will restore previous mocks.

test/utils/testHelper.ts
```diff
@@ -109,6 +135,17 @@ function mockAppConfigurationClientGetConfigurationSetting(kvList) {
     });
 }
 
+function mockAppConfigurationClientListConfigurationSettingsWithFailure(...pages: ConfigurationSetting[][]) {
+    const stub = sinon.stub(AppConfigurationClient.prototype, "listConfigurationSettings");
+
+    // Configure the stub to throw an error on the first call and return mockedKVs on the second call
+    stub.onFirstCall().throws(new RestError("Internal Server Error", { statusCode: 500 }));
```
zhiyuanliang-ms, line 142: Instead of stubing the listConfigurationSettings method for the `AppConfigurationClient.prototype`,  it is better to stub the client instances you created in mockGetClients.

test/failover.test.ts
```diff
@@ -0,0 +1,115 @@
...
+    });
+
+    it("should failover to replica and load key values from config store", async () => {
+        const replicaDiscoveryEnabled = true;
+        const isFailoverable = true;
+        mockConfigurationManagerGetClients(isFailoverable, mockedKVs);
+
+        const connectionString = createMockedConnectionString();
+        const settings = await load(connectionString, {
+            replicaDiscoveryEnabled: replicaDiscoveryEnabled
```
zhiyuanliang-ms, line 43: This should be enabled by default, isn't it?

zhiyuanliang-ms (approved):

zhiyuanliang-ms:
Looks good to me.

src/ConfigurationClientManager.ts
zhiyuanliang-ms: Unresolved.

zhiyuanliang-ms (changes requested):
List shuffling should be extracted to a function in utils.ts

zhiyuanliang-ms (approved):

src/AzureAppConfigurationImpl.ts
```diff
@@ -794,3 +828,9 @@ function getValidFeatureFlagSelectors(selectors?: SettingSelector[]): SettingSel
         return getValidSelectors(selectors);
     }
 }
+
+function isFailoverableError(error: any): boolean {
+    // ENOTFOUND: DNS lookup failed, ENOENT: no such file or directory
+    return isRestError(error) && (error.code === "ENOTFOUND" || error.code === "ENOENT" ||
```
zhiyuanliang-ms, line 834: Forget to handle `AuthenticationError`

zhiyuanliang-ms, line 834: https://nodejs.org/api/errors.html#common-system-errors Missed some error type

Issue #97, opened 2024-09-14
Package-lock.json is out of date


zhiyuanliang-ms:
There are two vulnerabilities in the package-lock. And the lockfileVersion is 2, after npm 8, the lockfileVersion should be 3.

Upgrade to typescript 5.6 will break the AzureAppConfigurationImpl because the ReadonlyMap API changed.

Eskibear:
`lockfileVersion` doesn't block you from fixing the vulnerabilities. Three tasks mentioned above, suggest to do them separately. 

> Upgrade to typescript 5.6  will break ...  the ReadonlyMap API changed.

Interesting it's not following SemVer. Better go thru the TS 5.6 release notes to figure out what's changed first.

PR #100, opened 2024-09-18
Update lockfile version & Audit vulnerable and reprecated packages

zhiyuanliang-ms:
## Why this PR?

#97 

The current npm version should be v10. The lockfile version 2 should be used for npm v6 or earlier.


zhiyuanliang-ms:
Move to #112

PR #101, opened 2024-09-19
Feature Flag Telemetry Support

zhiyuanliang-ms:
## Why this PR?

Populate feature flag id/feature flag reference/etag in feature flag's telemetry.metadata to facilitate Split experimentation.

Follows [PR](https://github.com/Azure/azure-sdk-for-python/pull/35254) and [PR](https://github.com/Azure/AppConfiguration-DotnetProvider/pull/517)

### DIFF
--- a/src/AzureAppConfigurationImpl.ts
+++ b/src/AzureAppConfigurationImpl.ts
@@ -258,8 +258,7 @@ export class AzureAppConfigurationImpl implements AzureAppConfiguration {
     }
 
     async #loadFeatureFlags() {
-        // Temporary map to store feature flags, key is the key of the setting, value is the raw value of the setting
-        const featureFlagsMap = new Map<string, any>();
+        const featureFlagSettings: ConfigurationSetting[] = [];
         for (const selector of this.#featureFlagSelectors) {
             const listOptions: ListConfigurationSettingsOptions = {
                 keyFilter: `${featureFlagPrefix}${selector.keyFilter}`,
@@ -276,7 +275,7 @@ export class AzureAppConfigurationImpl implements AzureAppConfiguration {
                 pageEtags.push(page.etag ?? "");
                 for (const setting of page.items) {
                     if (isFeatureFlag(setting)) {
-                        featureFlagsMap.set(setting.key, setting);
+                        featureFlagSettings.push(setting);
                     }
                 }
             }
@@ -285,7 +284,7 @@ export class AzureAppConfigurationImpl implements AzureAppConfiguration {
 
         // parse feature flags
         const featureFlags = await Promise.all(
-            Array.from(featureFlagsMap.values()).map(setting => this.#parseFeatureFlag(setting))
+            featureFlagSettings.map(setting => this.#parseFeatureFlag(setting))
         );
 
         // feature_management is a reserved key, and feature_flags is an array of feature flags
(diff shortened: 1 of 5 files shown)


src/AzureAppConfigurationImpl.ts
```diff
@@ -532,6 +537,83 @@ export class AzureAppConfigurationImpl implements AzureAppConfiguration {
         }
         return response;
     }
+
+    async #parseFeatureFlag(setting: ConfigurationSetting<string>): Promise<any> {
```
linglingye001, line 541: why async, same for line 561

src/AzureAppConfigurationImpl.ts
```diff
@@ -532,6 +537,83 @@ export class AzureAppConfigurationImpl implements AzureAppConfiguration {
...
+        }
+        // In Node.js, use the crypto module's hash function
+        else {
+            const hash = crypto.createHash("sha256").update(data).digest();
+            return hash.toString("base64url");
+        }
+    }
+
+    #createFeatureFlagReference(setting: ConfigurationSetting<string>): string {
+        let featureFlagReference = `${this.#clientEndpoint}kv/${setting.key}`;
```
linglingye001, line 611: Is there any requirement to featureFlagReference? what if in failover scenario when current endpoint from replica?

linglingye001, line 611: what if the client endpoint is undefined or CDN endpoints in the future?

src/AzureAppConfigurationImpl.ts
zhiyuanliang-ms: Line 597

https://stackoverflow.com/questions/57626477/using-javascript-crypto-subtle-in-synchronous-function

src/AzureAppConfigurationImpl.ts
zhiyuanliang-ms: Please read these two PRs: [.NET provider](https://github.com/Azure/AppConfiguration-DotnetProvider/pull/517) and [python provider](https://github.com/Azure/azure-sdk-for-python/pull/35254)

zhiyuanliang-ms: > what if in failover scenario when current endpoint from replica

Then we will use the replica url

zhiyuanliang-ms: Then, it will be the CDN endpoint.

examples/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "dependencies": {
-    "@azure/app-configuration-provider": "latest",
+    "@azure/app-configuration-provider": "../",
```
zhiyuanliang-ms, line 3: revert this change

examples/helloworld.mjs
```diff
@@ -19,8 +19,16 @@ const settings = await load(connectionString, {
     selectors: [{
         keyFilter: "app.settings.*"
     }],
-    trimKeyPrefixes: ["app.settings."]
+    trimKeyPrefixes: ["app.settings."],
```
zhiyuanliang-ms, line 22: revert this change

src/AzureAppConfigurationImpl.ts
linglingye001, line 611: Does the casing of "endpoint" affect subsequent hash calculation?

linglingye001 (approved):

rossgrambo (approved):

PR #106, opened 2024-10-11
New API to load from CDN endpoint

zhiyuanliang-ms:


src/load.ts
```diff
@@ -78,6 +78,54 @@ export async function load(
     }
 }
 
+/**
+ * Loads the data from a CDN and returns an instance of AzureAppConfiguration.
+ * @param endpoint  The URL to the CDN.
+ * @param options  Optional parameters.
+ */
+export async function loadCdn(endpoint: URL | string, options?: AzureAppConfigurationOptions): Promise<AzureAppConfiguration>;
```
zhenlan, line 86: I'm wondering if it makes more sense to name it `loadFromCDN` insead.

Eskibear: +1

src/load.ts
```diff
@@ -78,6 +78,54 @@ export async function load(
...
+            endpoint = new URL(endpoint);
+        } catch (error) {
+            if (error.code === "ERR_INVALID_URL") {
+                throw new Error("Invalid endpoint URL.", { cause: error });
+            } else {
+                throw error;
+            }
+        }
+    }
+    const emptyTokenCredential: TokenCredential = {
```
avanigupta, line 104: I think we can just call `load(endpoint, emptyTokenCredential, appConfigOptions)` without any extra validations here. `load` takes care of everything else.

src/load.ts
zhiyuanliang-ms: I will use camelCase, so it will be `loadFromCdn`

src/load.ts
```diff
@@ -78,6 +78,23 @@ export async function load(
...
+ * @param options  Optional parameters.
+ */
+export async function loadFromCdn(endpoint: URL | string, options?: AzureAppConfigurationOptions): Promise<AzureAppConfiguration>;
+
+export async function loadFromCdn(
+    endpoint: string | URL,
+    appConfigOptions?: AzureAppConfigurationOptions
+): Promise<AzureAppConfiguration> {
+    const emptyTokenCredential: TokenCredential = {
+        getToken: async () => ({ token: "", expiresOnTimestamp: 0 })
```
avanigupta, line 93: do we need to override the sync version?

zhiyuanliang-ms: image

The `TokenCredential` interface requires getToken to be async

avanigupta (approved):

src/load.ts
```diff
@@ -78,6 +78,23 @@ export async function load(
...
 
+/**
+ * Loads the data from a CDN and returns an instance of AzureAppConfiguration.
+ * @param endpoint  The URL to the CDN.
+ * @param options  Optional parameters.
+ */
+export async function loadFromCdn(endpoint: URL | string, options?: AzureAppConfigurationOptions): Promise<AzureAppConfiguration>;
+
+export async function loadFromCdn(
+    endpoint: string | URL,
```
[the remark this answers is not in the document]
zhiyuanliang-ms: I am ok with `cdnEndpoint`. Updated.

src/load.ts
```diff
@@ -82,6 +82,23 @@ export async function load(
     }
 }
 
+/**
+ * Loads the data from a CDN and returns an instance of AzureAppConfiguration.
+ * @param cdnEndpoint  The URL to the CDN.
+ * @param appConfigOptions  Optional parameters.
+ */
+export async function loadFromCdn(endpoint: URL | string, options?: AzureAppConfigurationOptions): Promise<AzureAppConfiguration>;
```
zhenlan, line 90: > endpoint

Should this be updated too?

zhiyuanliang-ms: weird, I remembered I updated this in the commit, will update it again

zhiyuanliang-ms:
Verified Azure Frond Door + sas token can get kvs from app config. So merge this PR.

PR #112, opened 2024-10-21
Update typescript version

zhiyuanliang-ms:
## Why this change

There is a breaking change in 5.6.x: https://devblogs.microsoft.com/typescript/announcing-typescript-5-6/#what’s-new-since-the-beta-and-rc 

The interface of `ReadonlyMap` changed. Our `AzureAppConfiguration` implements [it](https://github.com/Azure/AppConfiguration-JavaScriptProvider/blob/main/src/AzureAppConfiguration.ts#L19).

You may see some read underscore in vscode, this is because the built-in typescript version is 5.6+

## Visible change

Actually, there won't be visible change for users who consume our package. The compiled file will be written in javascript and there will be no types there. So the compiled outputs will be the same.


zhiyuanliang-ms:
> LGTM to update accordingly if it's a breaking change of ReadOnlyMap declaration from TS 5.6.x. No problem for JS developers. But for TS developers who are using TS Compiler < 5.6.x, can they compile the code after the change? If not, we need to add TS 5.6.x as pre-requisites.

I updated the dependency in package.json. We required the minimum version to 5.6.3

linglingye001 (approved):

PR #123, opened 2024-11-04
Load From Azure Front Door

zhiyuanliang-ms:
## Why this PR?

This feature is targeted on the CDN story.

This PR adds a new API `loadFromAzureFrontDoor` which allows user to load key values from the Azure Front Door.  The Azure Front Door will forward the request to App Config and use managed identity for authentication.

For CDN scenario, the recommended way for dynamic refresh is "watch all". We introduced key value collection monitoring based refresh in #133 .

The implementation is aligned with [#601](https://github.com/Azure/AppConfiguration-DotnetProvider/pull/601)

## Visible changes

No conditional request will be sent when CDN is used.

To break cdn cache, we will use a pipeline policy to append query param `_=<last-changed-etag>` to all the request sending to cdn.

Example:
```
Initial load - no etag

First watch request - no etag

Change detected:

1. sentinel key change 

All the following refresh/watch request (until new change detected) - sentinel key etag

2. sentinel key deleted

All the following refresh/watchrequest (until new change detected) - SHA256(ResourceDeleted\nprevious sentinel key etag) (Use hash to hide the "ResourceDeleted")

3. page etag change

All the following refresh/watch request (until new change detected) - the first changed page etag

We send different requests to retrieve feature flag and key values
```


juniwang (approved):

zhiyuanliang-ms:
I reverted the original commit of adding loadFromCdn (#130) to unblock doing release for EXP telemetry changes. 
We still have the issue with supporting dynamic refresh for CDN scenario. This PR will contains all changes of the complete solution of loading from cdn.

src/load.ts
```diff
@@ -82,6 +83,45 @@ export async function load(
...
+            const url = new URL(request.url);
+            url.searchParams.set("api-version", apiVersion);
+            request.url = url.toString();
+            return next(request);
+        },
+    };
+    if (appConfigOptions === undefined) {
+        appConfigOptions = { clientOptions: {}};
+    }
+    const policies = appConfigOptions.clientOptions?.additionalPolicies || [];
```
rossgrambo, line 116: We might want an additional 

```
if (appConfigOptions.clientOptions == null) {
    appConfigOptions.clientOptions = {}
}
```

Otherwise the policy wont be added if they provided an appConfigOptions with a null clientOptions

src/load.ts
```diff
@@ -82,6 +83,45 @@ export async function load(
...
+    // The policy position should be perRetry so that the policy will be processed after the api version policy added by the SDK.
+    // https://learn.microsoft.com/en-us/dotnet/api/azure.core.httppipelineposition?view=azure-dotnet#fields
+    policies.push({policy: apiVersionPolicy, position: "perRetry"});
+    appConfigOptions.clientOptions = { ...appConfigOptions.clientOptions, additionalPolicies: policies};
```
rossgrambo, line 120: Is it bad that this is a new object? If a developer passed in a clientOptions, constructed this, then updated something about the options, it would not reflect here. Not saying that's a common scenario but if the sdk allows for that we probably should too.

rossgrambo (approved):

src/load.ts
zhiyuanliang-ms: >  If a developer passed in a clientOptions, constructed this, then updated something about the options, it would not reflect here.

It doesn't sound like a valid use case. The options should be determined when you call the function since you passed the options as a parameter.

zhiyuanliang-ms:
Changed the target branch to the key value collection monitoring PR's branch #133 , as it will be the recommended way of dynamic refresh for CDN scenario.

rossgrambo:
This is blocked for now pending some CDN decisions

src/AzureAppConfigurationImpl.ts
```diff
@@ -974,57 +1049,48 @@ export class AzureAppConfigurationImpl implements AzureAppConfiguration {
...
             // Only use the first 15 bytes
             const first15Bytes = hash.slice(0, 15);
-
             return first15Bytes.toString("base64url");
         }
     }
+
+    async #calculateResourceDeletedCacheConsistencyToken(etag: string): Promise<string> {
+        const crypto = getCryptoModule();
+        const rawString = `ResourceDeleted\n${etag}`;
```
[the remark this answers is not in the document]
zhiyuanliang-ms: Not necessary. See the existing code in azure js sdk: [ref](https://github.com/Azure/azure-sdk-for-js/blob/fe622fde91562da5ebbc1dc49f162ff49ed35bb4/sdk/translation/ai-translation-text-rest/src/authenticationCustomized.ts#L29)

rossgrambo (approved):

src/AzureAppConfigurationImpl.ts
```diff
@@ -686,31 +730,62 @@ export class AzureAppConfigurationImpl implements AzureAppConfiguration {
...
+                    // if CDN is not used, add page etags to the listOptions to send conditional request
+                    listOptions = {
+                        ...listOptions,
+                        pageEtags: selector.pageEtags
+                    };
+                } else if (selectorCollection.cdnToken) {
+                    // If CDN is used, add etag to request header so that the pipeline policy can retrieve and append it to the request URL
+                    listOptions = {
+                        ...listOptions,
+                        requestOptions: { customHeaders: { [CDN_TOKEN_LOOKUP_HEADER]: selectorCollection.cdnToken } }
```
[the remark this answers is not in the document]
zhiyuanliang-ms: Added

PR #133, opened 2024-11-18
Refresh key value collection based on page etag

zhiyuanliang-ms:
## Why this PR?
This PR supports refreshing key value based on page etag (aka key value collection monitoring).

## Visible change
- When refreshOption.enabled is true and no watched setting is specified, then the provider will watch all selected key value to refresh the configuration

- This PR also re-organize  the `AzureAppConfigurationImpl` (1. adjust the method order 2. add more comments 3. simplify the code path 4. make more function resuable

## Note
Some corner cases:
1. If you load the following kvs and ffs:
``` javascript
{
  selector: [{keyFilter: "*"}],
  refreshOptions: {
    enabled: true
  }
}
```
with this load option, the `feature_management` section will not appear in the local configuration (provider will ignore feature flag configuration settings), because user doesn't use the feature flag in the load options (i.e. `AzureAppConfigurationOptions.feaetureFlagOptions`.

However, if users changes any feature flag through the portal, the page etag of monitored key value collection (of selector with keyFilter of "*")  will change. So it will trigger a real refresh. Even if, all the key value loaded to local configuration will be the same as the previous one.

This behavior may only affect the scenario when `onRefresh` callback is used. Otherwise, user will never notice this.

## Reference
.NET provider [PR](https://github.com/Azure/AppConfiguration-DotnetProvider/pull/574)

### DIFF
--- a/src/AzureAppConfigurationImpl.ts
+++ b/src/AzureAppConfigurationImpl.ts
@@ -63,6 +63,7 @@ export class AzureAppConfigurationImpl implements AzureAppConfiguration {
     #isFailoverRequest: boolean = false;
 
     // Refresh
+    #watchAll: boolean = false;
     #refreshInterval: number = DEFAULT_REFRESH_INTERVAL_IN_MS;
     #onRefreshListeners: Array<() => any> = [];
     /**
@@ -75,8 +76,13 @@ export class AzureAppConfigurationImpl implements AzureAppConfiguration {
     #featureFlagRefreshInterval: number = DEFAULT_REFRESH_INTERVAL_IN_MS;
     #featureFlagRefreshTimer: RefreshTimer;
 
-    // selectors
+    /**
+     * selectors of key-values obtained from @see AzureAppConfigurationOptions.selectors
+     */
     #keyValueSelectors: PagedSettingSelector[] = [];
+    /**
+     * selectors of feature flags obtained from @see AzureAppConfigurationOptions.featureFlagOptions.selectors
+     */
     #featureFlagSelectors: PagedSettingSelector[] = [];
 
     constructor(
@@ -94,25 +100,21 @@ export class AzureAppConfigurationImpl implements AzureAppConfiguration {
         }
 
         if (options?.refreshOptions?.enabled) {
-            const { watchedSettings, refreshIntervalInMs, watchAll } = options.refreshOptions;
-            // validate refresh options
-            if (watchAll !== true) {
-                if (watchedSettings === undefined || watchedSettings.length === 0) {
-                    throw new Error("Refresh is enabled but no watched settings are specified.");
-                } else {
-                    for (const setting of watchedSettings) {
-                        if (setting.key.includes("*") || setting.key.includes(",")) {
-                            throw new Error("The characters '*' and ',' are not supported in key of watched settings.");
-                        }
-                        if (setting.label?.includes("*") || setting.label?.includes(",")) {
-                            throw new Error("The characters '*' and ',' are not supported in label of watched settings.");
-                        }
-                        this.#sentinels.push(setting);
+            const { refreshIntervalInMs, watchedSettings } = options.refreshOptions;
+            if (watchedSettings === undefined || watchedSettings.length === 0) {
+                this.#watchAll = true; // if no watched settings is specified, then watch all
+            } else {
+                for (const setting of watchedSettings) {
+                    if (setting.key.includes("*") || setting.key.includes(",")) {
+                        throw new Error("The characters '*' and ',' are not supported in key of watched settings.");
                     }
+                    if (setting.label?.includes("*") || setting.label?.includes(",")) {
+                        throw new Error("The characters '*' and ',' are not supported in label of watched settings.");
+                    }
+                    this.#sentinels.push(setting);
                 }
-            } else if (watchedSettings && watchedSettings.length > 0) {
-                throw new Error("Watched settings should not be specified when registerAll is enabled.");
             }
+
             // custom refresh interval
             if (refreshIntervalInMs !== undefined) {
                 if (refreshIntervalInMs < MIN_REFRESH_INTERVAL_IN_MS) {
@@ -150,6 +152,27 @@ export class AzureAppConfigurationImpl implements AzureAppConfiguration {
         this.#adapters.push(new JsonKeyValueAdapter());
     }
 
+    get #refreshEnabled(): boolean {
+        return !!this.#options?.refreshOptions?.enabled;
+    }
+
+    get #featureFlagEnabled(): boolean {
+        return !!this.#options?.featureFlagOptions?.enabled;
+    }
+
+    get #featureFlagRefreshEnabled(): boolean {
+        return this.#featureFlagEnabled && !!this.#options?.featureFlagOptions?.refresh?.enabled;
+    }
+
+    get #requestTraceOptions() {
+        return {
+            requestTracingEnabled: this.#requestTracingEnabled,
+            initialLoadCompleted: this.#isInitialLoadCompleted,
+            appConfigOptions: this.#options,
+            isFailoverRequest: this.#isFailoverRequest
+        };
+    }
+
     // #region ReadonlyMap APIs
     get<T>(key: string): T | undefined {
         return this.#configMap.get(key);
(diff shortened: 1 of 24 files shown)


src/RefreshOptions.ts
```diff
@@ -24,6 +24,11 @@ export interface RefreshOptions {
      * Any modifications to watched settings will refresh all settings loaded by the configuration provider when refresh() is called.
      */
     watchedSettings?: WatchedSetting[];
+
+    /**
+     * Specifies whether all configuration settings will be watched for changes on the server.
+     */
+    watchAll?: boolean;
```
zhiyuanliang-ms, line 31: I prefer `watchAll` tp `registerAll`. In .NET, we have `RefreshOptions.Register` method. But, in javascript, we let user to list watchedSettings in the refreshOption.

zhenlan: Please don't follow the .NET design. When RefreshOptions.enabled is true but no individual keys are watched, it's watching all.

avanigupta: Just like .NET provider, if sentinel key is not provided, JS provider will also throw an error today. Are we saying that we stop throwing that error and assume that no sentinel means all selected kv are watched?

zhenlan: Note that .NET doesn't have an explicit option to enable refresh. The Register* API serves two purposes: 1. enable refresh 2. answer what to monitor. This is not the case with other providers. We have an explicit option to enable refresh in all other providers. This is why we don't need anything else in other providers to monitoring all.

src/AzureAppConfigurationImpl.ts
```diff
@@ -93,35 +94,38 @@ export class AzureAppConfigurationImpl implements AzureAppConfiguration {
...
+                            throw new Error("The characters '*' and ',' are not supported in key of watched settings.");
+                        }
+                        if (setting.label?.includes("*") || setting.label?.includes(",")) {
+                            throw new Error("The characters '*' and ',' are not supported in label of watched settings.");
+                        }
+                        this.#sentinels.push(setting);
+                    }
+                }
+            } else if (watchedSettings && watchedSettings.length > 0) {
+                throw new Error("Watched settings should not be specified when registerAll is enabled.");
```
linglingye001, line 114: should it be `watchAll`?

src/AzureAppConfigurationImpl.ts
```diff
@@ -93,35 +94,38 @@ export class AzureAppConfigurationImpl implements AzureAppConfiguration {
...
         if (options?.refreshOptions?.enabled) {
-            const { watchedSettings, refreshIntervalInMs } = options.refreshOptions;
-            // validate watched settings
-            if (watchedSettings === undefined || watchedSettings.length === 0) {
-                throw new Error("Refresh is enabled but no watched settings are specified.");
+            const { watchedSettings, refreshIntervalInMs, watchAll } = options.refreshOptions;
+            // validate refresh options
+            if (watchAll !== true) {
+                if (watchedSettings === undefined || watchedSettings.length === 0) {
+                    throw new Error("Refresh is enabled but no watched settings are specified.");
```
linglingye001, line 101: It's expected to watchAll if user enabled refresh but not specified sentinel keys

src/AzureAppConfigurationImpl.ts
```diff
@@ -458,6 +443,9 @@ export class AzureAppConfigurationImpl implements AzureAppConfiguration {
 
         // try refresh if any of watched settings is changed.
         let needRefresh = false;
+        if (this.#watchAll) {
+            needRefresh = await this.#checkKeyValueCollectionChanged(this.#keyValueSelectors);
+        }
         for (const sentinel of this.#sentinels.values()) {
```
linglingye001, line 449: What's the response when sentinels not existed in app config store?

src/AzureAppConfigurationImpl.ts
zhiyuanliang-ms: We will remove `watchAll` from the refreshOptions. See [here](https://github.com/Azure/AppConfiguration-JavaScriptProvider/pull/133#discussion_r1847440695)

src/AzureAppConfigurationImpl.ts
zhiyuanliang-ms: The sdk will throw REST exception with status code 404. However, the exception will be caught by the provider [here](https://github.com/Azure/AppConfiguration-JavaScriptProvider/blob/main/src/AzureAppConfigurationImpl.ts#L527)

src/AzureAppConfigurationImpl.ts
```diff
@@ -490,11 +478,20 @@ export class AzureAppConfigurationImpl implements AzureAppConfiguration {
...
+        const needRefresh = await this.#checkKeyValueCollectionChanged(this.#featureFlagSelectors);
+        if (needRefresh) {
+            await this.#loadFeatureFlags();
+        }
+
+        this.#featureFlagRefreshTimer.reset();
+        return Promise.resolve(needRefresh);
+    }
+
+    async #checkKeyValueCollectionChanged(selectors: PagedSettingSelector[]): Promise<boolean> {
```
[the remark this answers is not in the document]
linglingye001: In JS provider code implementation, there're many terms like `loadFeatureFlags`, `loadSelectedAndWatchedKeyValues`, `keyValueSelectors`, `featureFlagSelectors` ... are used for key values and feauture flags these two kinds settings.  Here `KeyValueCollection`  term make people confused what  kind of settings it checked. And it's strange when `featureFlagSelectors` are passed as parameter to this function. Besides, `ConfigurationSettings` is used in line239 when loaded setting can be both key values or feature flags. Let's keep good programming practice. Rename this function (like `checkConfiguraionSettingsChanged`) will bring our code base better Reusability, Readability, Maintainability.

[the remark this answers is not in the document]
zhiyuanliang-ms: `checkConfiguraionSettingsChanged` sounds good. Updated. 

The reason I used the name `key value collection` is because I refered .net provider
image

src/AzureAppConfigurationImpl.ts
```diff
@@ -514,30 +524,61 @@ export class AzureAppConfigurationImpl implements AzureAppConfiguration {
...
+                customOptions
+            );
+        };
+
+        let response: GetConfigurationSettingResponse | undefined;
+        try {
+            response = await this.#executeWithFailoverPolicy(funcToExecute);
+        } catch (error) {
+            if (isRestError(error) && error.statusCode === 404) {
+                response = undefined;
```
linglingye001, line 549: When customer enabled keyValue refresh and specified a non-existing sentinel, in this scenario it will never trigger a refresh right? Should we throw error or log warning to let the customer know the non-existing sentinel?

zhenlan: When the sentinel key is created, it will trigger a refresh. 404 is not an error case.

RichardChen820:
LGTM

src/AzureAppConfigurationImpl.ts
```diff
@@ -98,39 +105,38 @@ export class AzureAppConfigurationImpl implements AzureAppConfiguration {
...
-                    throw new Error("The characters '*' and ',' are not supported in label of watched settings.");
-                }
-                this.#sentinels.push(setting);
-            }
-
-            this.#refreshTimer = new RefreshTimer(this.#refreshInterval);
+            this.#kvRefreshTimer = new RefreshTimer(this.#kvRefreshInterval);
         }
 
+        this.#kvSelectors = getValidKeyValueSelectors(options?.selectors);
```
rossgrambo, line 134: Why does this KV selector happen outside the `if (options?.refreshOptions?.enabled) {` while the FF selector happens inside it's `if`?

src/AzureAppConfigurationImpl.ts
```diff
@@ -479,17 +336,141 @@ export class AzureAppConfigurationImpl implements AzureAppConfiguration {
...
     /**
-     * Refresh key-values.
+     * Loads configuration settings from App Configuration, either key-value settings or feature flag settings.
+     * Additionally, updates the `pageEtags` property of the corresponding @see PagedSettingSelector after loading.
+     *
+     * @param loadFeatureFlag - Determines which type of configurationsettings to load:
+     *                          If true, loads feature flag using the feature flag selectors;
+     *                          If false, loads key-value using the key-value selectors. Defaults to false.
+     */
+    async #loadConfigurationSettings(loadFeatureFlag: boolean = false): Promise<ConfigurationSetting[]> {
```
rossgrambo, line 346: nit: You could rewrite this function to get rid of the boolean flag. I think it's a little cleaner and more generic. It would look like:

```typescript
    async #loadConfigurationSettings(selectors: PagedSettingSelector[]): Promise<ConfigurationSetting[]> {
        const funcToExecute = async (client) => {
        ...
        }
        
        // Override selectors array with selectors to update array.
        selectors.splice(0, selectors.length, ...selectorsToUpdate);
        
        ...
```

test/refresh.test.ts
```diff
@@ -58,25 +58,6 @@ describe("dynamic refresh", function () {
         return expect(refreshCall).eventually.rejectedWith("Refresh is not enabled for key-values or feature flags.");
     });
 
-    it("should only allow non-empty list of watched settings when refresh is enabled", async () => {
```
rossgrambo, line 61: Do we want the inverse test? Should refresh all when no watched settings are provided

src/AzureAppConfigurationImpl.ts
zhiyuanliang-ms: Because we will always have key value selector, if user doesn't list any, we will use default selecter i.e. keyfilter="*", labelFilter=null. But for feature flag, only when user enabled feature flag, we will load it into configuration. This behavior is the same as .NET provider where you need to call `UseFeatureFlag`.

BTW, `if (options?.refreshOptions?.enabled)` branch checks the refresh option. I moved `this.#kvSelectors = getValidKeyValueSelectors(options?.selectors);` to the front of `if (options?.refreshOptions?.enabled)` and add more comments to make it more readable.

src/AzureAppConfigurationImpl.ts
zhiyuanliang-ms: I think we have to keep the boolean flag. The reason is that the selector can just be keyFilter="*" which will load all key-values includes feature flags. However, when featureFlagOption is not enabled, we should not load feature flag to the configuration. That's why I have these lines:

``` js
for (const setting of page.items) {
   if (loadFeatureFlag === isFeatureFlag(setting)) {
      loadedSettings.push(setting);
  }
}
```

src/AzureAppConfigurationImpl.ts
```diff
@@ -479,17 +337,141 @@ export class AzureAppConfigurationImpl implements AzureAppConfiguration {
...
+                };
+
+                const pageEtags: string[] = [];
+                const pageIterator = listConfigurationSettingsWithTrace(
+                    this.#requestTraceOptions,
+                    client,
+                    listOptions
+                ).byPage();
+                for await (const page of pageIterator) {
+                    pageEtags.push(page.etag ?? "");
```
amerjusupovic, line 369: What happens when the page etag is null and we just set it to empty string here?

zhiyuanliang-ms: AFAIK, page.etag shall not be null, even empty page should have a etag. But pageEtags's type is `string[]`. 
This is Yan's original implementation, I didn't change it: https://github.com/Azure/AppConfiguration-JavaScriptProvider/blob/main/src/AzureAppConfigurationImpl.ts#L275 @Eskibear

Eskibear: I don't remember clearly why but probably it's for simplicity, then we don't have to do null check later when comparing the etags. Alternatively, if `page.etag` can be null (from sdk declaration and implementation), we can also declare `pageEtags` as `(string | null)[]` or `Array<string | null>`.
After all, it depends on how we want to distinguish whether etag is null or empty. Is empty string a valid etag?

zhiyuanliang-ms: I don't think there will be anycase when etag is empty or null. Empty page should have a etag `00wWELzclWef22f82qcUDI7FbJweVOoKWZDA_JoUeD0`

avanigupta: Service will not respond with null or empty Etag (unless using older API versions).

amerjusupovic: Thanks for clarifying!

zhiyuanliang-ms: I will keep the current implementation.

src/AzureAppConfigurationImpl.ts
```diff
@@ -492,17 +346,141 @@ export class AzureAppConfigurationImpl implements AzureAppConfiguration {
...
+            if (matchedSetting) {
+                sentinel.etag = matchedSetting.etag;
+            } else {
+                // Send a request to retrieve key-value since it may be either not loaded or loaded with a different label or different casing
+                const { key, label } = sentinel;
+                const response = await this.#getConfigurationSetting({ key, label });
+                if (response) {
+                    sentinel.etag = response.etag;
+                } else {
+                    sentinel.etag = undefined;
```
linglingye001, line 436: can you make line436 consistent with line378? undefined or empty string

linglingye001 (approved):

src/AzureAppConfigurationImpl.ts
zhiyuanliang-ms: There is slight difference between two situations: line 436 is where we get 404 from the app config, line 378 is we get a page but the page.etag is undefined. Theoretically, this circumstance should not happen, discussion is [here](https://github.com/Azure/AppConfiguration-JavaScriptProvider/pull/133#discussion_r1873852763). But using undefined and "" can distinguish these two cases.

PR #135, opened 2024-11-19
Load balance support

linglingye001:


src/ConfigurationClientWrapper.ts
```diff
@@ -12,7 +12,7 @@ export class ConfigurationClientWrapper {
     endpoint: string;
     client: AppConfigurationClient;
     backoffEndTime: number = 0; // Timestamp
-    #failedAttempts: number = 0;
+    failedAttempts: number = 0;
```
zhiyuanliang-ms, line 15: If you have to expose failedAttempt to be public for testing, you should make this private and create another public property which only has getter. And the getter will return the private #failedAttempt. This can make sure failedAttempt won't be modified outside of the class.

src/ConfigurationClientWrapper.ts
```diff
@@ -21,11 +21,17 @@ export class ConfigurationClientWrapper {
 
     updateBackoffStatus(successfull: boolean) {
         if (successfull) {
-            this.#failedAttempts = 0;
+            if (this.failedAttempts > 0) {
+                this.failedAttempts = 0;
+            }
+            this.failedAttempts -= 1;
```
zhiyuanliang-ms, line 27: failedAttempts can be negative. This doesn't make sense. If this behavior is only for testing, don't do that.

test/loadBalance.test.ts
```diff
@@ -0,0 +1,113 @@
...
+                    keyFilter: "*"
+                }],
+                refresh: {
+                    enabled: true,
+                    refreshIntervalInMs: 2000 // 2 seconds for quick test.
+                }
+            }
+        });
+        // one request for key values, one request for feature flags
+        expect(fakeClientWrapper_1.failedAttempts).eq(-1);
```
zhiyuanliang-ms, line 56: Check this [pattern](https://github.com/Azure/AppConfiguration-JavaScriptProvider/blob/preview/test/refresh.test.ts#L243). You can follow this pattern to verify how many times something is called. You can do whatever you want when you stub the fakeClientWrapper. You can set a local counter and when the client is called, counter += 1

zhiyuanliang-ms (approved):

PR #136, opened 2024-11-28
Avoid multiple refresh

zhiyuanliang-ms:
## Why this PR?

#137

``` javascript
class Timer {
    #backoffEnd; // timestamp
    #interval;

    constructor(interval) {
        this.#interval = interval;
        this.#backoffEnd = Date.now();
    }

    canRefresh() {
        return Date.now() >= this.#backoffEnd;
    }

    reset() {
        this.#backoffEnd = Date.now() + this.#interval;
    }
}

async function refresh() {
    await f();
}

async function f() {
    if (!timer.canRefresh()) {
        console.log('timer rejects')
        return Promise.resolve(false);
    }
    console.log('f starts')
    await new Promise(resolve => setTimeout(resolve, 1000))
    console.log('f ends')
    timer.reset()
}

for (let i = 0; i < 3; i++) {
    refresh();
}
```

The above code snippet simulate the current implementation of refresh and it will produce the following output

```
f starts
f starts
f starts
f ends
f ends
f ends
```

The timer fails to prevent refresh operation from being executed multiple times during refresh interval.

## Solution

Except for the web worker scenario, javascript is in general single thread. For an async function, it will be synchronously executed until an await occurs, then it will give the control to the event loop. So this should be atomic. So the solution is to maintain a boolean variable `refreshInProgress` to indicate whether there is any ongoing refresh call. The check/setting of the flag will be done synchronously so it should be concurrent/"thread" safe.


src/common/lock.ts
```diff
@@ -0,0 +1,18 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT license.
+
+export class Lock {
```
Eskibear, line 4: Suggest to either use a less generic name or add some comments about the behavior. i.e. `execute` is aborted immediately if the it's locked instead of waiting until unlock.

RichardChen820:
Do we really need to address concurrency issue in this method? The `refresh` method's job is very simple, attempt to refresh config once, the concurrency issue really need to be considered by the caller depends on how the caller use it.

RichardChen820:
If we think it‘s of great value resolving it in the provider, I would rather like to have a new method something like `ExclusiveRefresh` to address it.

Eskibear:
> the concurrency issue really need to be considered by the caller depends on how the caller use it.

That makes sense. Then this task can simply become to add some details to the API mentioning [RefreshTimer](https://github.com/Azure/AppConfiguration-JavaScriptProvider/blob/main/src/refresh/RefreshTimer.ts), to make sure customers don't have the gap about the behavior.

https://github.com/Azure/AppConfiguration-JavaScriptProvider/blob/0027254e526a08ee42db970ed28c4cdfcce8349f/src/AzureAppConfiguration.ts#L6-L10

src/common/lock.ts
```diff
@@ -0,0 +1,18 @@
...
+    #locked = false;
```
RichardChen820, line 5: How we ensure the thread safety of this `locked` field?

zhiyuanliang-ms: Except for the web worker scenario, javascript is in general single thread. For an async function, it will be synchronously executed until an `await` occurs, then it will give the control to the event loop. So the operation
```
if (this.#locked) {
    return; // do nothing
}
this.#locked = true;
```
should be atomic.

linglingye001 (approved):

juniwang:
The latest code is much simpler, I like it!

PR #140, opened 2024-12-12
Support load snapshot

zhiyuanliang-ms:
## New API:

``` js
    selectors: [
        { snapshotName: "MySnapshot" }, // load keyvalue from snapshot
        { keyFilter: "test*", labelFilter: "test" }
    ],
    featureFlagOptions: {
        enabled: true,
        selectors: [
            { snapshotName: "MySnapshot" }, // load feature flags from snapshot
            { keyFilter: "*", labelFilter: "test" }
        ]
    }
});
```

Allow to use snapshot selector to load key value or feature flags

Note, for expilicity, if a snapshot contains feature flag, the feature flags will only be loaded when the snapshot is used for `featureFlagOptions.selectors`

This implementation is the same as kubernetes provider [ref](https://github.com/Azure/AppConfiguration-KubernetesProvider/blob/13b345feca1cf212a50aa77994693c32f6c4ae86/config/crd/bases/azconfig.io_azureappconfigurationproviders.yaml#L155) and [ref](https://github.com/Azure/AppConfiguration-KubernetesProvider/blob/13b345feca1cf212a50aa77994693c32f6c4ae86/api/v1/azureappconfigurationprovider_types.go#L79)


juniwang:
the impl looks good to me. Should we add or update our test cases to cover snapshot?

zhiyuanliang-ms:
@juniwang Do you mean add more testcases?

juniwang:
yes

src/AzureAppConfigurationImpl.ts
```diff
@@ -695,6 +741,29 @@ export class AzureAppConfigurationImpl implements AzureAppConfiguration {
...
+        let response: GetSnapshotResponse | undefined;
+        try {
+            response = await this.#executeWithFailoverPolicy(funcToExecute);
+        } catch (error) {
+            if (isRestError(error) && error.statusCode === 404) {
+                response = undefined;
+            } else {
+                throw error;
+            }
+        }
```
RichardChen820, line 763: Seems this try catch block is not so necessary, you handle the error in the caller anyway, can just handle the rest error 404 directly there?

zhiyuanliang-ms: I refered .NET provider's [implementation](https://github.com/Azure/AppConfiguration-DotnetProvider/blob/8b883cce448004c901bb9ef3a855ee6576bd2281/src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationProvider.cs#L885)

The `#getSnapshot` method has the same pattern as `#getConfigurationSetting` helper method. It just a wrapper of `ConfigurationClient.getSnapshot`. 

During the process of loading key value from snapshot, we will first get the snapshot itself to check whether it exists and whether its composition type is correct. I want to put this part of logic in the `loadConfigurationSettings` method to make it easier to follow.

Notice that we don't throw RestError with 404 directly, we throw InvalidOperationError with helpful error message instead.

RichardChen820: Wrap up everything in `getSnapshot` does make sense. Let's keep it

RichardChen820 (approved):

PR #153, opened 2025-01-15
Add feature management package version tracing

zhiyuanliang-ms:
## Why this PR?

Collect JS FM package version usage in request tracing

Note that there is an exported constant [`VERSION`](https://github.com/microsoft/FeatureManagement-JavaScript/blob/main/src/feature-management/src/index.ts#L8) in JS FM package. This is by design to be a hook for us to know the package version.

### DIFF
--- a/src/AzureAppConfigurationImpl.ts
+++ b/src/AzureAppConfigurationImpl.ts
@@ -66,8 +66,6 @@ export class AzureAppConfigurationImpl implements AzureAppConfiguration {
     #isInitialLoadCompleted: boolean = false;
     #isFailoverRequest: boolean = false;
     #featureFlagTracing: FeatureFlagTracingOptions | undefined;
-
-    #isPackageInspected: boolean = false;
     #fmVersion: string | undefined;
 
     // Refresh
@@ -231,7 +229,7 @@ export class AzureAppConfigurationImpl implements AzureAppConfiguration {
      * Loads the configuration store for the first time.
      */
     async load() {
-        await this.#ensurePackageInspected();
+        await this.#inspectFmPackage();
         await this.#loadSelectedAndWatchedKeyValues();
         if (this.#featureFlagEnabled) {
             await this.#loadFeatureFlags();
@@ -287,7 +285,6 @@ export class AzureAppConfigurationImpl implements AzureAppConfiguration {
      * Refreshes the configuration.
      */
     async refresh(): Promise<void> {
-        await this.#ensurePackageInspected();
         if (!this.#refreshEnabled && !this.#featureFlagRefreshEnabled) {
             throw new Error("Refresh is not enabled for key-values or feature flags.");
         }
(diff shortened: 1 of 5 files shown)


src/AzureAppConfigurationImpl.ts
```diff
@@ -316,6 +323,19 @@ export class AzureAppConfigurationImpl implements AzureAppConfiguration {
         return new Disposable(remove);
     }
 
+    async #ensurePackageInspected() {
+        if (!this.#isPackageInspected) {
+            this.#isPackageInspected = true;
+            try {
+                // get feature management package version
+                const fmPackage = await import(FM_PACKAGE_NAME);
```
zhiyuanliang-ms, line 331: This line could cause webpack warning: `Critical dependency: the request of a dependency is an expression`



To suppress the warning, we can add this `/* webpackIgnore: true */` directive above this line.

zhiyuanliang-ms: I prefer to not add the webpack-specific directive here.

zhiyuanliang-ms:
@jimmyca15 @linglingye @zhenlan @rossgrambo @mrm9084 @juniwang 

I want to merge this PR before we stablize v2. Could you help review this PR please?

src/AzureAppConfigurationImpl.ts
```diff
@@ -316,6 +323,19 @@ export class AzureAppConfigurationImpl implements AzureAppConfiguration {
         return new Disposable(remove);
     }
 
+    async #ensurePackageInspected() {
+        if (!this.#isPackageInspected) {
+            this.#isPackageInspected = true;
```
linglingye001, line 328: Not sure FM package tracing applies for what scenarios. In your code, `isPackageInspected` will be enabled unconditionally. What if user disabled request tracing or not enabled featureFlags?

zhiyuanliang-ms: This is intended. The  inspectation of feature management package should only happen once.

zhiyuanliang-ms: Similar pattern in .net provide: [ref](https://github.com/Azure/AppConfiguration-DotnetProvider/blob/main/src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationProvider.cs#L29)

src/AzureAppConfigurationImpl.ts
```diff
@@ -66,6 +67,9 @@ export class AzureAppConfigurationImpl implements AzureAppConfiguration {
     #isFailoverRequest: boolean = false;
     #featureFlagTracing: FeatureFlagTracingOptions | undefined;
 
+    #isPackageInspected: boolean = false;
```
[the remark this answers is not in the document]
zhiyuanliang-ms: To make sure, we only inspect feature management package for once.

src/AzureAppConfigurationImpl.ts
zhiyuanliang-ms: Replied in this [thread](https://github.com/Azure/AppConfiguration-JavaScriptProvider/pull/153#discussion_r1944413241)

linglingye001: Got it,  I misread it earlier  and thought the `try` block was outside the `if` block.

linglingye001, line 331: shouldn't fmPackage import happens when requestTracing enabled? like [.net](https://github.com/Azure/AppConfiguration-DotnetProvider/blob/main/src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationProvider.cs#L1221)

zhiyuanliang-ms: good call. updated

linglingye001 (approved):

PR #154, opened 2025-01-16
Add express example

zhiyuanliang-ms:


examples/package.json
```diff
@@ -1,7 +1,8 @@
 {
   "dependencies": {
-    "@azure/app-configuration-provider": "latest",
+    "@azure/app-configuration-provider": "2.0.0-preview.2",
```
zhiyuanliang-ms, line 3: This is a transient change. Before I merge this PR, I will revert to latest. Ideally, we will stablize v2 after the lunar new year holiday.

examples/server.mjs
```diff
@@ -0,0 +1,43 @@
...
+
+import express from "express";
+
+const app = express();
+const PORT = 3000;
+
+app.use(express.json());
+
+app.get("/", (req, res) => {
+    appConfig.refresh();
```
linglingye001, line 27: Is there any middleware or handler that can be customized to call refresh() method?

examples/server.mjs
```diff
@@ -0,0 +1,48 @@
...
+});
+
+import express from "express";
+
+const server = express();
+const PORT = 3000;
+
+server.use(express.json());
+
+// Use a middleware to achieve request-driven configuration refresh
```
zhiyuanliang-ms, line 27: It should reference our public doc once this [PR](https://github.com/MicrosoftDocs/azure-docs-pr/pull/292628/files) is merged

examples/server.mjs
```diff
@@ -0,0 +1,48 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT license.
+
+import * as dotenv from "dotenv";
+dotenv.config()
+
+import { load } from "@azure/app-configuration-provider";
+const connectionString = process.env.APPCONFIG_CONNECTION_STRING;
```
zhiyuanliang-ms, line 8: use token credential

examples/refresh.mjs
```diff
@@ -26,20 +26,18 @@ const settings = await load(connectionString, {
...
-
     // Current value of message
     console.log(settings.get("message"));
-
-    // Waiting before the next refresh
-    await sleepInMs(5000);
+    // Refreshing the configuration setting asynchronously
+    settings.refresh();
+    // wait for 5 seconds
+    await sleepInMs(5_000);
```
linglingye001, line 42: Any consideration for changes in this file?

zhiyuanliang-ms: The refresh operation should be non-block. I also updated the example in our public [doc](https://learn.microsoft.com/en-us/azure/azure-app-configuration/enable-dynamic-configuration-javascript?tabs=configuration-map)

If we await on refresh, it is against our design principle for dynamic refresh.

linglingye001: The request driven refresh applies to web scenario, however these description was put in console applications section in JS public doc, what the so-called activity to a console app?

zhiyuanliang-ms: I didn't mention request driven refresh in console example

examples/refresh.mjs
```diff
@@ -26,20 +26,18 @@ const settings = await load(connectionString, {
...
+        refreshIntervalInMs: 10_000 // Default value is 30 seconds, shorted for this sample
     }
 });
 
 console.log("Using Azure portal or CLI, update the `app.settings.message` value, and then update the `app.settings.sentinel` value in your App Configuration store.")
 
 // eslint-disable-next-line no-constant-condition
 while (true) {
-    // Refreshing the configuration setting
-    await settings.refresh();
```
zhenlan, line 38: This is a console app. It makes sense to keep the `await`. It will be helpful if we can update the comments to indicate this is a blocking call and users can remove `await` if they want it to be asynchronous. What matters is that users understand the implications and make conscious choices. There is nothing wrong with having `await`.

examples/server.mjs
zhenlan, line 1: Should we put different examples in different folders?

examples/server.mjs
```diff
@@ -0,0 +1,51 @@
...
+dotenv.config()
+
+import { load } from "@azure/app-configuration-provider";
+import { DefaultAzureCredential } from "@azure/identity";
+const endpoint = process.env.APPCONFIG_ENDPOINT;
+const credential = new DefaultAzureCredential();
+const appConfig = await load(endpoint, credential, {
+    refreshOptions: {
+        enabled: true,
+        refreshIntervalInMs: 5_000
```
zhenlan, line 14: Remove. People will copy/paste. Please don't put short refresh intervals in our examples.

examples/server.mjs
```diff
@@ -0,0 +1,51 @@
...
+server.get("/", (req, res) => {
+    res.send("Please go to /config to get the configuration.");
+});
+
+server.get("/config", (req, res) => {
+    res.json(appConfig.constructConfigurationObject());
+});
+
+server.get("/config/:key", (req, res) => {
+    res.json(appConfig.get(req.params.key) ?? "");
```
zhenlan, line 46: This is a weird example. Can we actually put up a page and use configuration for real?

examples/server.mjs
```diff
@@ -0,0 +1,51 @@
...
+const credential = new DefaultAzureCredential();
+const appConfig = await load(endpoint, credential, {
+    refreshOptions: {
+        enabled: true,
+        refreshIntervalInMs: 5_000
+    }
+});
+
+appConfig.onRefresh(() => {
+    console.log("Configuration has been refreshed.");
```
zhenlan, line 19: In a real example, we won't have a fake log but something like `appConfig.constructConfigurationObject()` here. That's how this callback is supposed to be used.

examples/server.mjs
zhiyuanliang-ms: How about simply return a message from App Config?

examples/console-example/configObject.mjs
```diff
@@ -1,45 +1,45 @@
...
+import { load } from "@azure/app-configuration-provider";
+const connectionString = process.env.APPCONFIG_CONNECTION_STRING;
+const settings = await load(connectionString, {
+    selectors: [{
+        keyFilter: "app.settings.*"
+    }]
+});
+
+const config = settings.constructConfigurationObject({
+    separator: "."
```
zhenlan, line 41: I thought `.` was the default separator so it's necessary. If you want to show case you can add a comment.

examples/web-example/httpserver.mjs
```diff
@@ -0,0 +1,51 @@
...
+// Use a middleware to achieve request-driven configuration refresh
+// For more information, please go to dynamic refresh tutorial: https://learn.microsoft.com/azure/azure-app-configuration/enable-dynamic-configuration-javascript
+server.use((req, res, next) => {
+    // this call is not blocking, the configuration will be updated asynchronously
+    appConfig.refresh();
+    next();
+});
+
+server.get("/", (req, res) => {
+    res.send(`Message from Azure App Configuration: ${appConfig.get("message")}`);
```
zhenlan, line 42: You took all the trouble to get the config object constructed. You should use the config object (`config.messsage`?).

examples/console-example/README.md
zhenlan, line 1: Everything is under the _examples_ folder. I don't think we need to add "example" to every child folder. How about _console-app_ and _web-app_?

src/common/disposable.ts
```diff
@@ -15,4 +15,4 @@ export class Disposable {
         }
         this.#disposed = true;
     }
-}
+}
```
zhenlan, line 18: What's changed for this file?

examples/web-example/httpserver.mjs
```diff
@@ -0,0 +1,51 @@
...
+});
+
+server.get("/config", (req, res) => {
```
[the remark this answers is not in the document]
zhiyuanliang-ms: This could be useful if another app/component want to fetch configuration from this server. I think it is a common scenario in JavaScript. I prefer to keep it.

examples/console-example/README.md
zhiyuanliang-ms: Sounds good.

src/common/disposable.ts
zhiyuanliang-ms: Add an extra empty line at the end of the file.

examples/web-example/httpserver.mjs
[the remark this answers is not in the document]
zhenlan: It's a legit scenario to build a web API service. Users can also use appconfig to manage the behavior of their service. But what's the scenario they expose all their configuration? Why do we think this is an important scenario to show? Why do we believe users cannot figure out how to do it assuming there is legit reason of doing it?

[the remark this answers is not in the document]
zhiyuanliang-ms: Removed

examples/console-app/refresh.mjs
```diff
@@ -1,45 +1,42 @@
...
+    }
+});
+
+console.log("Using Azure portal or CLI, update the `app.settings.message` value, and then update the `app.settings.sentinel` value in your App Configuration store.")
+
+// eslint-disable-next-line no-constant-condition
+while (true) {
+    console.log(settings.get("message"));
+    // this is a blocking call and you can remove await to make the refresh operation asynchronous
+    await settings.refresh();
```
zhenlan, line 39: Scenario-wise, it makes more sense to put `await settings.refresh()` before the `settings.get("message")`. This means you are using the latest configuration.

zhenlan (approved):

examples/console-app/refresh.mjs
zhiyuanliang-ms: done

Issue #157, opened 2025-02-03
Comment of featureFlagOptions doesn't match the real behavior


zhiyuanliang-ms:
https://github.com/Azure/AppConfiguration-JavaScriptProvider/blob/main/src/featureManagement/FeatureFlagOptions.ts#L22

https://github.com/Azure/AppConfiguration-JavaScriptProvider/blob/main/src/AzureAppConfigurationImpl.ts#L626

Comment: https://github.com/Azure/AppConfiguration-JavaScriptProvider/pull/65#discussion_r1626639561

zhiyuanliang-ms:
JS provider will throw exception if no selectors is specified for feature flags instead of load all feature flags with null label.

This is inconsistent with all of our other providers.

Discussed offline with @zhenlan . We should make it consistent.

PR #158, opened 2025-02-06
Load all feature flag with no label when there is no selector specified

zhiyuanliang-ms:
#157


zhiyuanliang-ms:
This PR need to be merged before stalizing v2

linglingye001 (approved):

PR #166, opened 2025-02-17
Startup Retry & Configurable Startup Time-out & Error handling

zhiyuanliang-ms:
## Why this PR?

- Support startup retry and timeout.
- Handle error in a more managed way
- Added boot loop protection [ref](https://github.com/Azure/AppConfiguration-DotnetProvider/blob/main/src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationProvider.cs#L174)

The retry/backoff behavior is the same as .NET provider.

Retriable error in .NET provider: [ref](https://github.com/Azure/AppConfiguration-DotnetProvider/blob/main/src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationProvider.cs#L664)

reference: [#488](https://github.com/Azure/AppConfiguration-DotnetProvider/pull/488)

## Usage

```javascript
  startupOptions: { 
    timeoutInMs: 30_000
});
```

By default the startup timeout will be 100 seconds.

The pseudo code of the whole startup retry + failover logic:

```
function startupWithRetry():
    while not timeout:
        try:
            result = loadWithFailover()
            return result  // Load successful; exit the retry loop
        except error:
            if isRetriable(error):
                backoffDuration = calculateBackoffDuration()
                sleep(backoffDuration)  // Wait for the backoff duration before retrying
            else:
                throw error  // Non-retriable error; rethrow immediately

function loadWithFailover():
    // Retrieve the fallback client list via DNS SRV records
    fallbackClients = getFallbackClientsFromDNS()
    
    // Attempt to perform the load operation using each fallback client sequentially
    for client in fallbackClients:
        try:
            result = client.load()
            return result  // Load successful; return result immediately
        except error:
            if not isFailoverable(error):
                // If the error is not failoverable, immediately rethrow the error
                throw error
            // If the error is failoverable, continue with the next client
    
    // If all fallback clients have failed, throw a retriable error to trigger another retry
    throw new Error("All fallback clients failed")
```


zhenlan:
Is there a reason the retry can be disabled?

zhiyuanliang-ms:
> Is there a reason the retry can be disabled?

If people want to throw immediatly when initial load fails, they can disable it. Setting a very short timeout is not feasible for this case. The `retryEnabled` is true by default. So, there is no behavior difference from other providers.

zhenlan:
> If people want to...

Anyone asked for it? What's the scenario for it?

zhiyuanliang-ms:
> > If people want to...
> 
> Anyone asked for it? What's the scenario for it?

No one asks for it. I personally want to disable the retry so that I can know whether there is something wrong immediately. 
I have experienced this: I used a wrong connection string (with correct format but wrong content), the `load` call lasted for 100 sec before fail. Same problem exists in .NET provider.

zhenlan:
> No one asks for it. I personally want to disable the retry so that I can know whether there is something wrong immediately.

If the connection string is malformed, I hope it will throw so we can fail immediately. Otherwise, you can't really tell what's wrong. Set a shorter timeout if you don't want to wait for too long.

> are you worrying about that even if the default behavior is retry-enabled , user will still choose to disable it and we don't want them to disable it in most cases? Otherwise, I don't see why we cannot keep it.

Configuration is in critical code path. Most applications can't start without configuration loaded properly. Transient errors can happen for any cloud solutions. We rely on the client retry (and failover etc.) to deliver customers high availability. Please don't design a "feature" so customers can shoot their own feet.

src/error.ts
```diff
@@ -0,0 +1,38 @@
...
+    constructor(message: string) {
+        super(message);
+        this.name = "FailoverError";
+    }
+}
+
+export function isFailoverableError(error: any): boolean {
+    // ENOTFOUND: DNS lookup failed, ENOENT: no such file or directory
+    return isRestError(error) && (error.code === "ENOTFOUND" || error.code === "ENOENT" ||
+        (error.statusCode !== undefined && (error.statusCode === 401 || error.statusCode === 403 || error.statusCode === 408 || error.statusCode === 429 || error.statusCode >= 500)));
```
RichardChen820, line 29: I would suggest rewrite the checks for readability 

```
if (!isRestError(error)){
   return false
}

if (error.code == "ENOTFOUND" || error.code === "ENOENT"){
   return true;
}

if (error.StatusCode ===undefined) {
   return false;
}

if ( error.statusCode === 401 || ... || ...||...) {
   return true;
}

return false;

```

src/AzureAppConfigurationImpl.ts
```diff
@@ -639,7 +697,7 @@ export class AzureAppConfigurationImpl implements AzureAppConfiguration {
         }
 
         this.#clientManager.refreshClients();
-        throw new Error("All clients failed to get configuration settings.");
+        throw new FailoverError("All fallback clients failed to get configuration settings.");
```
RichardChen820, line 700: From my perspective, there's no failover error concept at all. This error just indicates it fails to initiate or refresh after trying every possible client. And callers even get nothing special to do with a `FailoverError`

src/ConfigurationClientManager.ts
```diff
@@ -58,7 +65,7 @@ export class ConfigurationClientManager {
                 this.#id = regexMatch[2];
                 this.#secret = regexMatch[3];
             } else {
-                throw new Error(`Invalid connection string. Valid connection strings should match the regex '${ConnectionStringRegex.source}'.`);
+                throw new RangeError(`Invalid connection string. Valid connection strings should match the regex '${ConnectionStringRegex.source}'.`);
```
RichardChen820, line 68: In the case, call it as a RangeError sounds odd, how about rename it to `ArgumentError` or `ParameterError`?

zhiyuanliang-ms: [RangeError](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/RangeError) is a built-in error type in JavaScript. There is no built-in ArgumentError. I did consider about create an ArgumentError type by myself. But I think it overlaps with `RangeError`

src/error.ts
zhiyuanliang-ms: Sounds good.

src/failover.ts
```diff
@@ -0,0 +1,39 @@
...
+    if (timeElapsed <= 200_000) { // 200 seconds in milliseconds
+        return 10_000; // 10 seconds in milliseconds
+    }
+    if (timeElapsed <= 10 * 60 * 1000) { // 10 minutes in milliseconds
+        return MIN_BACKOFF_DURATION;
+    }
+    return undefined;
+}
+
+export function calculateDynamicBackoffDuration(failedAttempts: number) {
```
RichardChen820, line 23: Not sure what the `dynamic` word refers to?

src/AzureAppConfigurationImpl.ts
```diff
@@ -320,6 +340,44 @@ export class AzureAppConfigurationImpl implements AzureAppConfiguration {
...
+                try {
+                    await this.#loadSelectedAndWatchedKeyValues();
+                    if (this.#featureFlagEnabled) {
+                        await this.#loadFeatureFlags();
+                    }
+                    // Mark all settings have loaded at startup.
+                    this.#isInitialLoadCompleted = true;
+                    break;
+                } catch (error) {
+
```
RichardChen820, line 360: nit: extra line.

src/failover.ts
zhiyuanliang-ms: To distinguish it with the fixed backoff

src/failover.ts
```diff
@@ -0,0 +1,39 @@
...
+    if (failedAttempts <= 1) {
+        return MIN_BACKOFF_DURATION;
+    }
+
+    // exponential: minBackoff * 2 ^ (failedAttempts - 1)
+    const exponential = Math.min(failedAttempts - 1, MAX_SAFE_EXPONENTIAL);
+    let calculatedBackoffDuration = MIN_BACKOFF_DURATION * (1 << exponential);
```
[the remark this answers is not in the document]
zhiyuanliang-ms: We've handled it by MAX_SAFE_EXPONENTIAL. But you are right, Math.pow() is better.

src/AzureAppConfigurationImpl.ts
```diff
@@ -320,6 +340,44 @@ export class AzureAppConfigurationImpl implements AzureAppConfiguration {
...
+                    if (!isRetriableError(error)) {
+                        throw error;
+                    }
+                    if (abortSignal.aborted) {
+                        return;
+                    }
+                    const timeElapsed = Date.now() - startTimestamp;
+                    let postAttempts = 0;
```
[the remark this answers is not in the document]
zhiyuanliang-ms: Refer to https://github.com/Azure/AppConfiguration-DotnetProvider/blob/main/src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationProvider.cs#L617

test/requestTracing.test.ts
```diff
@@ -35,15 +35,25 @@ describe("request tracing", function () {
...
         expect(headerPolicy.headers.get("User-Agent")).satisfy((ua: string) => ua.startsWith("javascript-appconfiguration-provider"));
     });
 
     it("should have request type in correlation-context header", async () => {
         try {
-            await load(createMockedConnectionString(fakeEndpoint), { clientOptions });
+            await load(createMockedConnectionString(fakeEndpoint), {
+                clientOptions,
+                startupOptions: {
+                    timeoutInMs: 1
```
linglingye001, line 54: Why set the timeout to 1 ms and not use the default time out?

src/StartupOptions.ts
```diff
@@ -0,0 +1,14 @@
...
+// Licensed under the MIT license.
+
+export const DEFAULT_STARTUP_TIMEOUT_IN_MS = 100 * 1000; // 100 seconds in milliseconds
+
+export interface StartupOptions {
+    /**
+     * The amount of time allowed to load data from Azure App Configuration on startup.
+     *
+     * @remarks
+     * If not specified, the default value is 100 seconds. Must be greater than 1 second.
```
linglingye001, line 11: Is there any validation or check on this when user set timeout parameter?

src/keyvault/AzureKeyVaultKeyValueAdapter.ts
```diff
@@ -24,7 +25,7 @@ export class AzureKeyVaultKeyValueAdapter implements IKeyValueAdapter {
     async processKeyValue(setting: ConfigurationSetting): Promise<[string, unknown]> {
         // TODO: cache results to save requests.
         if (!this.#keyVaultOptions) {
-            throw new Error("Configure keyVaultOptions to resolve Key Vault Reference(s).");
+            throw new OperationError("Failed to process the key vault reference. The keyVaultOptions is not configured.");
```
zhiyuanliang-ms, line 28: RangeError

src/keyvault/AzureKeyVaultKeyValueAdapter.ts
```diff
@@ -43,7 +44,7 @@ export class AzureKeyVaultKeyValueAdapter implements IKeyValueAdapter {
             return [setting.key, await this.#keyVaultOptions.secretResolver(new URL(sourceId))];
         }
 
-        throw new Error("No key vault credential or secret resolver callback configured, and no matching secret client could be found.");
```
zhiyuanliang-ms, line 46: Catch error during getSecret().

src/AzureAppConfigurationImpl.ts
```diff
@@ -286,7 +306,7 @@ export class AzureAppConfigurationImpl implements AzureAppConfiguration {
      */
     async refresh(): Promise<void> {
         if (!this.#refreshEnabled && !this.#featureFlagRefreshEnabled) {
-            throw new Error("Refresh is not enabled for key-values or feature flags.");
+            throw new OperationError("Refresh is not enabled for key-values or feature flags.");
```
zhiyuanliang-ms, line 309: Should be RangeError

src/AzureAppConfigurationImpl.ts
```diff
@@ -320,6 +340,43 @@ export class AzureAppConfigurationImpl implements AzureAppConfiguration {
...
+                    let postAttempts = 0;
+                    let backoffDuration = getFixedBackoffDuration(timeElapsed);
+                    if (backoffDuration === undefined) {
+                        postAttempts += 1;
+                        backoffDuration = calculateBackoffDuration(postAttempts);
+                    }
+                    await new Promise(resolve => setTimeout(resolve, backoffDuration));
+                    console.warn("Failed to load configuration settings at startup. Retrying...");
+                }
+            } while (!abortSignal.aborted);
```
linglingye001, line 376: Since the check on abortSignal is at the end of each try operation, so there may be a scenario that when it's still in loading process but timeout happens, program still need to run a little while before abort. May be you can setup timeout for signal like this https://developer.mozilla.org/en-US/docs/Web/API/AbortSignal/timeout_static

test/requestTracing.test.ts
zhiyuanliang-ms: Most `load` calls in requestTracing testsuit will always thrown RESTError with code ENOTFOUND (which is retriable error). Now we have startup retry, I don't want to have each testcase last for 100 sec

src/StartupOptions.ts
zhiyuanliang-ms: I forget to remove the last sentence in comment. I used to implement this constraint, but I remove it as .NET provider doesn't have it.

src/AzureAppConfigurationImpl.ts
zhiyuanliang-ms: I did think about this option, it is an alternative for using Promise.race for timeout.

If we take this option, the code will be

```javascript
async function load() {
	try {
		await initializeWithRetryPolicy(AbortSignal.timeout(5200)); // 200ms longer than default backoff duration
	} catch (error) {}
}

async function initializeWithRetryPolicy(signal) {
	do {
		try {
			//--> Position A
			// load keyvalue from App Config
		} catch (error) {
		 	//--> Position B
			// wait for 5000ms (default backoff duration)
		}
	} while (signal.abort)

	if (signal.abort) {
		throw new Error("Timeout")
	}
}
```
Let's say one failure of load kv from App Config (position A) costs 50ms
At T0, we set timeout for the signal (At T0+5200, it should be aborted)
At T0 + 50ms, the first attempt failed and it went to position B and started to wait for 5000ms
At T0+5050ms, it started to retry and it took 50m to fail again
At T0+5100ms, failed at position A again and started to wait for 5000ms or maybe even longer due to exponential backoff duration increase
At T0+5200ms, the signal timeout, but the code is still stuck at await initialize.


> program still need to run a little while before abort.

This is half-true. Unless the customer write code like this:
``` javascript
try {
	const setting = await load();
} catch (err) {
	// do not throw here
}
```
The initializeWithRetryPolicy promise will not be resolved immediately after timeout. Otherwise, error is thrown, the node process is terminated. All promises running no longer exist.

src/ConfigurationClientManager.ts
avanigupta: Looks like JS doesnt have anything equivalent to ArgumentError. By definition, RangeError is also not a good fit here. I'd vote for custom error that extends `Error` class, or simply throw `Error`.

src/AzureAppConfigurationImpl.ts
```diff
@@ -320,6 +340,42 @@ export class AzureAppConfigurationImpl implements AzureAppConfiguration {
...
+                        return;
+                    }
+                    const timeElapsed = Date.now() - startTimestamp;
+                    let postAttempts = 0;
+                    let backoffDuration = getFixedBackoffDuration(timeElapsed);
+                    if (backoffDuration === undefined) {
+                        postAttempts += 1;
+                        backoffDuration = calculateBackoffDuration(postAttempts);
+                    }
+                    console.warn(`Failed to load. Error message: ${error.message}. It Will retry in ${backoffDuration} ms.`);
```
amerjusupovic, line 372:
```suggestion
                    console.warn(`Failed to load. Error message: ${error.message}. It will retry in ${backoffDuration} ms.`);

```

amerjusupovic: I was trying to make this closer to [.NET provider logs](https://github.com/Azure/AppConfiguration-DotnetProvider/blob/main/src/Microsoft.Extensions.Configuration.AzureAppConfiguration/LogHelper.cs) that are based on .NET SDK logs, I'm not sure if javascript SDK logs look the same?

src/keyvault/AzureKeyVaultKeyValueAdapter.ts
```diff
@@ -70,6 +79,6 @@ export class AzureKeyVaultKeyValueAdapter implements IKeyValueAdapter {
     }
 }
 
-function getHost(url: string) {
-    return new URL(url).host;
+function buildKeyVaultReferenceErrorMessage(message: string, setting: ConfigurationSetting, secretIdentifier?: string ): string {
+    return `${message} Key: ${setting.key} Label: ${setting.label ?? ""} ETag: ${setting.etag ?? ""} SecretIdentifier: ${secretIdentifier ?? ""}`;
```
amerjusupovic, line 83:
```suggestion
    return `${message} Key: '${setting.key}' Label: '${setting.label ?? ""}' ETag: '${setting.etag ?? ""}' SecretIdentifier: '${secretIdentifier ?? ""}'`;

```

src/AzureAppConfigurationImpl.ts
linglingye001: When timeout, the `TimeoutError` will be thrown by the abort signal.

zhiyuanliang-ms: @linglingye001 Have you tried the code snippet? TimeoutError will only be thrown if the signal is used in `fetch`.

The signal returned by AbortSignal.timeout(time) will only:
image

Please run some simple code snippet in Node.js first.

zhiyuanliang-ms: In browser:
image

In node:
```js
const signal = AbortSignal.timeout(1)
let i = 0
while(i < 100) {
    console.log("Yes")
    console.log(signal.aborted)
    i += 1
}
console.log(signal)
```
image

Reason:
image

linglingye001: Discussed offline.

src/error.ts
```diff
@@ -0,0 +1,70 @@
...
+    if (error instanceof ArgumentError ||
+        error instanceof OperationError ||
+        error instanceof TypeError ||
+        error instanceof RangeError) {
+        return false;
+    }
+    return true;
+}
+
+export function isInstantlyThrowError(error: any): boolean {
```
avanigupta, line 63: What's the difference between isRetriableError and isInstantlyThrowError? The error types are the same except for one. Do we need this distinction?

src/error.ts
```diff
@@ -0,0 +1,70 @@
...
+        this.name = "KeyVaultReferenceError";
+    }
+}
+
+export function isFailoverableError(error: any): boolean {
+    if (!isRestError(error)) {
+        return false;
+    }
+    // ENOTFOUND: DNS lookup failed, ENOENT: no such file or directory
+    if (error.code == "ENOTFOUND" || error.code === "ENOENT") {
```
avanigupta, line 41: Can we check if `error.code !== undefined`

avanigupta, line 41: Why is one condition with `==` comparator and the other with `===` comparator?

src/error.ts
```diff
@@ -0,0 +1,70 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT license.
+
+import { isRestError } from "@azure/core-rest-pipeline";
+
+/**
+ * Error thrown when an operation cannot be performed by the Azure App Configuration provider.
+ */
+export class OperationError extends Error {
```
avanigupta, line 9: Is this more like InvalidOperation or OperationCancelled?

src/common/utils.ts
```diff
@@ -2,9 +2,25 @@
...
-  for (let i = array.length - 1; i > 0; i--) {
-    const j = Math.floor(Math.random() * (i + 1));
-    [array[i], array[j]] = [array[j], array[i]];
-  }
-  return array;
+    for (let i = array.length - 1; i > 0; i--) {
+        const j = Math.floor(Math.random() * (i + 1));
+        [array[i], array[j]] = [array[j], array[i]];
+    }
+    return array;
```
avanigupta, line 9: Why the extra space?

src/common/utils.ts
```diff
@@ -2,9 +2,25 @@
...
+}
+
+export function getEndpointUrl(endpoint: string): URL {
+    try {
+        return new URL(endpoint);
```
avanigupta, line 14: Creating a new URL can only throw TypeError. No need to catch and rethrow as TypeError. 
That means you can directly return new URL(endpoint).
In fact, this helper method is not needed.

src/common/utils.ts
```diff
@@ -2,9 +2,25 @@
...
+    } catch (error) {
+        throw new TypeError(`Invalid Endpoint URL: ${endpoint}`);
+    }
+}
+
+export function getUrlHost(url: string): string {
```
avanigupta, line 20: Helper method not needed.
As a general guideline, if the helper method is just a simple line of code, it's probably not worth the redirection.

src/failover.ts
```diff
@@ -0,0 +1,38 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT license.
+
+const MIN_BACKOFF_DURATION = 30_000; // 30 seconds in milliseconds
+const MAX_BACKOFF_DURATION = 10 * 60 * 1000; // 10 minutes in milliseconds
+const JITTER_RATIO = 0.25;
+
+// The backoff duration algorithm is consistent with the .NET configuration provider's implementation.
```
avanigupta, line 8: This comment is not very helpful in understanding what fixed backoff is. You could instead add an explanation of what this method is doing.

src/refresh/refreshOptions.ts
avanigupta, line 1: File name should be PascalCase

zhiyuanliang-ms: In javascript, camel case is more common. If you go to see the vscode code base ([ref](https://github.com/microsoft/vscode/tree/main/src/vs/editor/common)) and azure js sdk code base ([ref](https://github.com/Azure/azure-sdk-for-js/tree/main/sdk/appconfiguration/app-configuration/src))

I also explained it in this [thread](https://github.com/microsoft/FeatureManagement-JavaScript/pull/93#discussion_r1952059099)

The issue that our current JS library is using a mixed file name style existed when I took over the repos from @Eskibear. 
I will fix it. It is on my backlog.

src/error.ts
zhiyuanliang-ms: I followed .net provider's [implementation](https://github.com/Azure/AppConfiguration-DotnetProvider/blob/main/src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationProvider.cs#L170)

The difference is that whether we need to wait for an additional delay. How about renaming the method name to `isArgumentError`?

src/error.ts
zhiyuanliang-ms: This is the implementation of `isRestError` from azure js sdk:
image

If error is `undefined`, the code will not enter line 41.

Besides, we only call `isXXXError` in `catch(error){ ... }` branch where error can never be undefined.

zhiyuanliang-ms: It is a mistake. Good catch!

src/error.ts
zhiyuanliang-ms: Renamed to `InvalidOperationError`

src/common/utils.ts
zhiyuanliang-ms: which space?

src/common/utils.ts
zhiyuanliang-ms: This method was introduced in this [PR](https://github.com/Azure/AppConfiguration-JavaScriptProvider/pull/98/files#diff-d56b8d0c484133ecb5a8a9fd31502d676cc0df5b566f80be4b230d3757772312R273)

But I agree with that this method is unnecessary.

src/common/utils.ts
zhiyuanliang-ms: Removed.

src/failover.ts
zhiyuanliang-ms: Removed this comment.

src/failover.ts
```diff
@@ -0,0 +1,37 @@
...
+const JITTER_RATIO = 0.25;
+
+export function getFixedBackoffDuration(timeElapsed: number): number | undefined {
+    if (timeElapsed <= 100_000) { // 100 seconds in milliseconds
+        return 5_000; // 5 seconds in milliseconds
+    }
+    if (timeElapsed <= 200_000) { // 200 seconds in milliseconds
+        return 10_000; // 10 seconds in milliseconds
+    }
+    if (timeElapsed <= 10 * 60 * 1000) { // 10 minutes in milliseconds
```
CsCherrYY, line 15: Should it be `timeElapsed <= MAX_BACKOFF_DURATION`?

src/refresh/refreshOptions.ts
```diff
@@ -1,44 +1,44 @@
...
-}
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT license.
+
+import { WatchedSetting } from "../WatchedSetting.js";
+
+export const DEFAULT_REFRESH_INTERVAL_IN_MS = 30 * 1000;
+export const MIN_REFRESH_INTERVAL_IN_MS = 1 * 1000;
+
+export interface RefreshOptions {
```
CsCherrYY, line 9: Why change the first character of the file name from upper case to lower case? Is there any rule to define the file names in this repository?

src/error.ts
```diff
@@ -0,0 +1,68 @@
...
+
+export function isRetriableError(error: any): boolean {
+    if (isArgumentError(error) ||
+        error instanceof RangeError) {
+        return false;
+    }
+    return true;
+}
+
+export function isArgumentError(error: any): boolean {
```
CsCherrYY, line 61: The method name looks a little ambiguous to me. Why method `isArgumentError` will return true when the error is `ArgumentError`, `TypeError` or `RangeError`? Does the first kind include the following kinds?

src/refresh/refreshOptions.ts
zhiyuanliang-ms: https://github.com/microsoft/FeatureManagement-JavaScript/pull/93#discussion_r1952059099

src/failover.ts
zhiyuanliang-ms: I copied the implementation of .net provider: https://github.com/Azure/AppConfiguration-DotnetProvider/blob/main/src/Microsoft.Extensions.Configuration.AzureAppConfiguration/Extensions/TimeSpanExtensions.cs#L18

CsCherrYY: I see, let's keep it as it is.

src/error.ts
CsCherrYY: If it's by design, I suggest you to add a comment for this method to describe the reason, to avoid other's confusion.

RichardChen820:
LGTM

RichardChen820 (approved):

src/failover.ts
```diff
@@ -0,0 +1,37 @@
...
+}
+
+export function calculateBackoffDuration(failedAttempts: number) {
+    if (failedAttempts <= 1) {
+        return MIN_BACKOFF_DURATION;
+    }
+
+    // exponential: minBackoff * 2 ^ (failedAttempts - 1)
+    // The right shift operator is not used in order to avoid potential overflow. Bitwise operations in JavaScript are limited to 32 bits.
+    let calculatedBackoffDuration = MIN_BACKOFF_DURATION * Math.pow(2, failedAttempts - 1);
```
avanigupta, line 28: This change is not related to the PR. Avoid modifying existing functionality which is not related to the feature being implemented. If its a bug or enhancement, it should be addressed in a separate PR.

src/failover.ts
```diff
@@ -0,0 +1,37 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT license.
+
+const MIN_BACKOFF_DURATION = 30_000; // 30 seconds in milliseconds
+const MAX_BACKOFF_DURATION = 10 * 60 * 1000; // 10 minutes in milliseconds
+const JITTER_RATIO = 0.25;
+
+export function getFixedBackoffDuration(timeElapsed: number): number | undefined {
+    if (timeElapsed <= 100_000) { // 100 seconds in milliseconds
```
avanigupta, line 9: .NET uses '<' comparison instead of '<='

https://github.com/Azure/AppConfiguration-DotnetProvider/blob/65ed48033482e796143b41fa150d06293fbf8316/src/Microsoft.Extensions.Configuration.AzureAppConfiguration/Extensions/TimeSpanExtensions.cs#L120-L128

src/StartupOptions.ts
```diff
@@ -0,0 +1,14 @@
...
+export const DEFAULT_STARTUP_TIMEOUT_IN_MS = 100 * 1000; // 100 seconds in milliseconds
+
+export interface StartupOptions {
+    /**
+     * The amount of time allowed to load data from Azure App Configuration on startup.
+     *
+     * @remarks
+     * If not specified, the default value is 100 seconds.
+     */
+    timeoutInMs?: number;
```
avanigupta, line 13: Can we assign the default value here so that we dont need to export DEFAULT_STARTUP_TIMEOUT_IN_MS

src/ConfigurationClientManager.ts
```diff
@@ -54,18 +59,18 @@ export class ConfigurationClientManager {
             const regexMatch = connectionString.match(ConnectionStringRegex);
             if (regexMatch) {
                 const endpointFromConnectionStr = regexMatch[1];
-                this.endpoint = getValidUrl(endpointFromConnectionStr);
+                this.endpoint = new URL(endpointFromConnectionStr);
```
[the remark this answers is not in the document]
zhiyuanliang-ms: The previous implementation of `getValidUrl` is 

```js
function getValidUrl(endpoint: string): URL {
    try {
        return new URL(endpoint);
    } catch (error) {
        if (error.code === "ERR_INVALID_URL") {
            throw new Error("Invalid endpoint URL.", { cause: error });
        } else {
            throw error;
        }
    }
}
```
image
If URL is invalid, new URL() will throw TypeError. I think getValidUrl method is not that useful. So I removed it.

src/ConfigurationClientManager.ts
```diff
@@ -272,20 +277,3 @@ function getClientOptions(options?: AzureAppConfigurationOptions): AppConfigurat
...
-    } catch (error) {
-        if (error.code === "ERR_INVALID_URL") {
-            throw new Error("Invalid endpoint URL.", { cause: error });
-        } else {
-            throw error;
-        }
-    }
-}
-
-export function instanceOfTokenCredential(obj: unknown) {
```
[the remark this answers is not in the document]
zhiyuanliang-ms: I have removed this method as it is kind of useless.

src/StartupOptions.ts
zhiyuanliang-ms: reverted the change

src/ConfigurationClientManager.ts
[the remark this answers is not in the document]
zhiyuanliang-ms: reverted the change

src/failover.ts
zhiyuanliang-ms: I moved this function from [src/ConfigurationClientWrapper.ts](https://github.com/Azure/AppConfiguration-JavaScriptProvider/pull/166/files#diff-c9b0738545def7e2bd4ac955cd8d841d21502f725fab968f5db28956318581b2) I want to reorganize the project structure.

I added a new file called `failover.ts`, I put all functions related to backoff there. I renamed the `failover.ts` to `backoffDuration.ts`. Let it make more sense. Would that be alright with you?

src/StartupOptions.ts
zhiyuanliang-ms: No, interface doesn't allow to have default value for properties. If we want to have default value, then we have to use Class.
Then, user will lose the convenience to write code like this:
``` js
const config = await load("xxx", { startOptions: { timeoutInMs: 10_000} });
```
```js
They must write code like this const config = await load("xxx", { startOptions: new StartopOptions(10_000) });
```
It is kind of anti-javascript convention. You can see that using interface for options is a very common practice in azure js [sdk](https://github.com/Azure/azure-sdk-for-js/blob/024894c79fc334b0269f10fadb53ba2d98edce87/sdk/appconfiguration/app-configuration/src/models.ts#L19).

src/failover.ts
avanigupta: I was referring to the change to use Math.pow instead of bitwise operation. While refactoring code, try not to make any changes to the internal logic of the methods being moved around.

src/ConfigurationClientManager.ts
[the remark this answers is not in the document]
avanigupta: We should handle such errors and surface them in the form of ArgumentError. This is an invalid input scenario and should be handled like other invalid input errors.

src/keyvault/AzureKeyVaultKeyValueAdapter.ts
```diff
@@ -24,33 +25,41 @@ export class AzureKeyVaultKeyValueAdapter implements IKeyValueAdapter {
...
+            // precedence: secret clients > credential > secret resolver
+            const client = this.#getSecretClient(new URL(vaultUrl));
+            if (client) {
+                const secret = await client.getSecret(secretName, { version });
+                return [setting.key, secret.value];
+            }
+            if (this.#keyVaultOptions.secretResolver) {
+                return [setting.key, await this.#keyVaultOptions.secretResolver(new URL(sourceId))];
+            }
+        } catch (error) {
```
avanigupta, line 44: We cannot catch every error and throw as KeyVaultReferenceError. We need to handle only those errors that are expected from Key Vault. See this for .NET error handling: https://github.com/Azure/AppConfiguration-DotnetProvider/blob/bf8b06b9dde1bb219b625939b3d2ea00bd2a61d5/src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureKeyVaultReference/AzureKeyVaultKeyValueAdapter.cs#L49-L56

src/keyvault/AzureKeyVaultKeyValueAdapter.ts
```diff
@@ -24,33 +25,41 @@ export class AzureKeyVaultKeyValueAdapter implements IKeyValueAdapter {
     async processKeyValue(setting: ConfigurationSetting): Promise<[string, unknown]> {
         // TODO: cache results to save requests.
         if (!this.#keyVaultOptions) {
-            throw new Error("Configure keyVaultOptions to resolve Key Vault Reference(s).");
+            throw new ArgumentError("Failed to process the key vault reference. The keyVaultOptions is not configured.");
         }
 
-        // precedence: secret clients > credential > secret resolver
         const { name: secretName, vaultUrl, sourceId, version } = parseKeyVaultSecretIdentifier(
```
avanigupta, line 31: Can parseKeyVaultSecretIdentifier utility from KeyVault SDK throw an error? If yes, we need to catch it and throw KeyVaultReferenceError.

https://github.com/Azure/AppConfiguration-DotnetProvider/blob/bf8b06b9dde1bb219b625939b3d2ea00bd2a61d5/src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureKeyVaultReference/AzureKeyVaultKeyValueAdapter.cs#L38-L41

src/keyvault/AzureKeyVaultKeyValueAdapter.ts
```diff
@@ -24,33 +25,41 @@ export class AzureKeyVaultKeyValueAdapter implements IKeyValueAdapter {
...
+            const client = this.#getSecretClient(new URL(vaultUrl));
+            if (client) {
+                const secret = await client.getSecret(secretName, { version });
+                return [setting.key, secret.value];
+            }
+            if (this.#keyVaultOptions.secretResolver) {
+                return [setting.key, await this.#keyVaultOptions.secretResolver(new URL(sourceId))];
+            }
+        } catch (error) {
+            throw new KeyVaultReferenceError(buildKeyVaultReferenceErrorMessage(error.message, setting, sourceId));
```
avanigupta, line 45: Is there a way to preserve the stack trace of this error when creating a new KeyVaultReferenceError?
For eg, in .NET, we preserve the exception from KeyVault by setting it as inner exception of KeyVaultReferenceException: 

https://github.com/Azure/AppConfiguration-DotnetProvider/blob/bf8b06b9dde1bb219b625939b3d2ea00bd2a61d5/src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureKeyVaultReference/AzureKeyVaultKeyValueAdapter.cs#L66C24-L66C66

src/keyvault/AzureKeyVaultKeyValueAdapter.ts
```diff
@@ -24,33 +25,41 @@ export class AzureKeyVaultKeyValueAdapter implements IKeyValueAdapter {
     async processKeyValue(setting: ConfigurationSetting): Promise<[string, unknown]> {
         // TODO: cache results to save requests.
         if (!this.#keyVaultOptions) {
-            throw new Error("Configure keyVaultOptions to resolve Key Vault Reference(s).");
+            throw new ArgumentError("Failed to process the key vault reference. The keyVaultOptions is not configured.");
```
avanigupta, line 28:
```suggestion
            throw new ArgumentError("Failed to process the Key Vault reference because Key Vault options are not configured.");

```

avanigupta: Avoid hardcoding method/variable names in text. If method name changes in future, the error message will become invalid. Is there a JS alternative for C# `nameof` operator?

src/backoffDuration.ts
```diff
@@ -0,0 +1,37 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT license.
+
+const MIN_BACKOFF_DURATION = 30_000; // 30 seconds in milliseconds
+const MAX_BACKOFF_DURATION = 10 * 60 * 1000; // 10 minutes in milliseconds
+const JITTER_RATIO = 0.25;
+
+export function getFixedBackoffDuration(timeElapsed: number): number | undefined {
```
avanigupta, line 8: This method has too many comments which make it harder to read. Suggest changing input argument name to `timeElapsedInMs` to clarify that we're dealing with milliseconds. That way, no comments are needed.

src/error.ts
```diff
@@ -0,0 +1,70 @@
...
+ */
+export class ArgumentError extends Error {
+    constructor(message: string) {
+        super(message);
+        this.name = "ArgumentError";
+    }
+}
+
+/**
+ * Error thrown when it fails to get the secret from the Key Vault.
```
avanigupta, line 27:
```suggestion
 * Error thrown when a Key Vault reference cannot be resolved.
```

src/error.ts
```diff
@@ -0,0 +1,70 @@
...
+ */
+export class KeyVaultReferenceError extends Error {
+    constructor(message: string) {
+        super(message);
+        this.name = "KeyVaultReferenceError";
+    }
+}
+
+export function isFailoverableError(error: any): boolean {
+    if (!isRestError(error)) {
```
avanigupta, line 37: What is rest error?

src/error.ts
```diff
@@ -0,0 +1,70 @@
...
+        return false;
+    }
+    // ENOTFOUND: DNS lookup failed, ENOENT: no such file or directory
```
avanigupta, line 40: We also need to check for transient network errors:
https://github.com/Azure/AppConfiguration-DotnetProvider/blob/bf8b06b9dde1bb219b625939b3d2ea00bd2a61d5/src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationProvider.cs#L1253-L1256

src/error.ts
```diff
@@ -0,0 +1,70 @@
...
+
+    return false;
+}
+
+export function isRetriableError(error: any): boolean {
+    if (isArgumentError(error)) {
+        return false;
+    }
+    return true;
+}
```
avanigupta, line 58: isRetriableError is the negation of isArgumentError. We can skip adding this method and directly call `!isArgumentError` wherever needed

src/backoffDuration.ts
```diff
@@ -0,0 +1,37 @@
...
+    if (timeElapsed < 200_000) { // 200 seconds in milliseconds
+        return 10_000; // 10 seconds in milliseconds
+    }
+    if (timeElapsed < 10 * 60 * 1000) { // 10 minutes in milliseconds
+        return MIN_BACKOFF_DURATION;
+    }
+    return undefined;
+}
+
+export function calculateBackoffDuration(failedAttempts: number) {
```
avanigupta, line 21: Since the other method is called `getFixedBackoffDuration`, this can be renamed to `getExponentialBackoffDuration`

src/error.ts
```diff
@@ -0,0 +1,70 @@
...
+    // 401 Unauthorized, 403 Forbidden, 408 Request Timeout, 429 Too Many Requests, 5xx Server Errors
+    if (error.statusCode !== undefined &&
+        (error.statusCode === 401 || error.statusCode === 403 || error.statusCode === 408 || error.statusCode === 429 || error.statusCode >= 500)) {
+        return true;
+    }
+
+    return false;
+}
+
+export function isRetriableError(error: any): boolean {
```
avanigupta, line 53: Will InvalidOperationError be retriable?

src/error.ts
```diff
@@ -0,0 +1,70 @@
...
+    if (isArgumentError(error)) {
+        return false;
+    }
+    return true;
+}
+
+/**
+ * Check if the error is an instance of ArgumentError, TypeError, or RangeError.
+ */
+export function isArgumentError(error: any): boolean {
```
avanigupta, line 63:
```suggestion
export function isInputError(error: any): boolean {
```

src/error.ts
zhiyuanliang-ms: This is an error type from Azure JS SDK. Any failure related to http request will be treated as rest error.

src/error.ts
zhiyuanliang-ms: There is no such error in javascript. They will all be treated as RestError by sdk.

src/error.ts
zhiyuanliang-ms: Sounds good.

src/error.ts
zhiyuanliang-ms: InvalidOperationError is only be thrown in `constructConfigurationObject` method. This is out of retry path.

```js
const appConfig = await load("xxx");
const config = appConfig.constructConfigurationObject(); // <- InvalidOperationError will only be thrown here, if invalid separator is found
```

src/backoffDuration.ts
zhiyuanliang-ms: I like it!

src/error.ts
avanigupta: Is there no error code to identify if a RestError is a transient network error?
In that case I'm not sure I follow what the rest error check is doing above. 

If it's a rest error, only then we check the error code to say if it's failoverable. If it's not a rest error, it's never failoverable?

src/backoffDuration.ts
```diff
@@ -0,0 +1,37 @@
+// Copyright (c) Microsoft Corporation.
```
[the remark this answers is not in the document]
zhiyuanliang-ms: sounds good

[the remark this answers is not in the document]
zhiyuanliang-ms: image
I will name it `backoffUtils`

src/backoffDuration.ts
zhiyuanliang-ms: Great suggestion!

src/keyvault/AzureKeyVaultKeyValueAdapter.ts
zhiyuanliang-ms: > Is there a JS alternative for C# nameof operator?

No. Updated the error message.

src/keyvault/AzureKeyVaultKeyValueAdapter.ts
zhiyuanliang-ms:
```js
function parseKeyVaultSecretIdentifier(id) {
    const urlParts = id.split("/");
    const collection = urlParts[3];
    return Object.assign({ sourceId: id }, parseKeyVaultIdentifier(collection, id));
}
function parseKeyVaultIdentifier(collection, identifier) {
    if (typeof collection !== "string" || !(collection = collection.trim())) {
        throw new Error("Invalid collection argument");
    }
    if (typeof identifier !== "string" || !(identifier = identifier.trim())) {
        throw new Error("Invalid identifier argument");
    }
    let baseUri;
    try {
        baseUri = new URL(identifier);
    }
    catch (e) {
        throw new Error(`Invalid ${collection} identifier: ${identifier}. Not a valid URI`);
    }
    // Path is of the form '/collection/name[/version]'
    const segments = (baseUri.pathname || "").split("/");
    if (segments.length !== 3 && segments.length !== 4) {
        throw new Error(`Invalid ${collection} identifier: ${identifier}. Bad number of segments: ${segments.length}`);
    }
    if (collection !== segments[1]) {
        throw new Error(`Invalid ${collection} identifier: ${identifier}. segment [1] should be "${collection}", found "${segments[1]}"`);
    }
    const vaultUrl = `${baseUri.protocol}//${baseUri.host}`;
    const name = segments[2];
    const version = segments.length === 4 ? segments[3] : undefined;
    return {
        vaultUrl,
        name,
        version,
    };
}
```
This is the implementation of `parseKeyVaultSecretIdentifier`.

> Can parseKeyVaultSecretIdentifier utility from KeyVault SDK throw an error? If yes, we need to catch it and throw KeyVaultReferenceError.

Make sense.

src/keyvault/AzureKeyVaultKeyValueAdapter.ts
zhiyuanliang-ms: It can be achieved through ErrorOptions which allows you to parse a cause.

Updated the KeyVaultReferenceError to take an error as its cause.

src/keyvault/AzureKeyVaultKeyValueAdapter.ts
zhiyuanliang-ms: The js sdk doesn't handle the error type with very fine-grained granularity. As far as I know, there is only [`RestError`](https://github.com/Azure/azure-sdk-for-js/blob/main/sdk/core/core-rest-pipeline/src/restError.ts#L37) and the general `Error` type.

But as you said [here](https://github.com/Azure/AppConfiguration-JavaScriptProvider/pull/166#discussion_r2059044367), we should do error handling for parseKeyVaultSecretIdentifier. So I cannot only handle RestError.

zhiyuanliang-ms: I don't think it is necessary to check the properties of the error and provide different error message based on the erro type. I think using [ErrorOptions.cause](https://github.com/Azure/AppConfiguration-JavaScriptProvider/pull/166#discussion_r2059556034) to wrap the inner error could be a decent solution. Because the full information of the inner exception will be provided in the error message. User can tell what is going on from it.

zhiyuanliang-ms: There are only [RestError](https://github.com/Azure/azure-sdk-for-js/blob/main/sdk/core/core-rest-pipeline/src/restError.ts#L37) , [AuthenticationError](https://github.com/Azure/azure-sdk-for-js/blob/main/sdk/identity/identity/src/errors.ts#L95) and the general Error type.

I just found that we missed handling `AuthenticationError` in the failover [PR](https://github.com/Azure/AppConfiguration-JavaScriptProvider/pull/98).

src/error.ts
zhiyuanliang-ms: image

src/failover.ts
zhiyuanliang-ms: I discussed this with @Eskibear offline. Math.pow is generally better than the bitwise operation. I can create a seprate PR.

zhiyuanliang-ms: #189

src/ConfigurationClientManager.ts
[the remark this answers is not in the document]
zhiyuanliang-ms: Now TypeError will be handled in the `isInputError`. 

So you are suggesting that even if the a TypeError is thrown, we should catch it and wrap it as an ArgumentError? I think it is unnecessary.

src/keyvault/AzureKeyVaultKeyValueAdapter.ts
zhiyuanliang-ms: As you said [here](https://github.com/Azure/AppConfiguration-JavaScriptProvider/pull/166#discussion_r2059044367), we should also handle error from `parseKeyVaultSecretIdentifier`. But only general `Error` type will be thrown there.

src/error.ts
zhiyuanliang-ms: Handled these error code in the PR.

src/common/error.ts
```diff
@@ -0,0 +1,64 @@
...
+ */
+export class KeyVaultReferenceError extends Error {
+    constructor(message: string, options?: ErrorOptions) {
+        super(message, options);
+        this.name = "KeyVaultReferenceError";
+    }
+}
+
+export function isFailoverableError(error: any): boolean {
+    if (!isRestError(error)) {
```
avanigupta, line 37: Does this mean that only RestError is failoverable? What kind of errors can we expect from JS SDK which may not be RestError?

src/keyvault/AzureKeyVaultKeyValueAdapter.ts
avanigupta: > I have a question why `UnauthorizedAccessException` is not treated as [failoverable](https://github.com/Azure/AppConfiguration-DotnetProvider/blob/bf8b06b9dde1bb219b625939b3d2ea00bd2a61d5/src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationProvider.cs#L1224) error in .net provider.

Are you asking about unauthorized error from key vault? None of the errors from KeyVault are failoverable. The reason is that failing over to a different AppConfig replica will not help with any KeyVault error.

avanigupta: Its unfortunate that KeyVault SDK throws generic Error with no error code for parseKeyVaultSecretIdentifier.
I like your suggestion. Wrap both KeyVault calls in different try/catch blocks and handle the errors separately.

src/ConfigurationClientManager.ts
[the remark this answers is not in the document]
avanigupta: Oh you mean `new URL()` will throw `TypeError` instead of `Error`?
Then we dont need to wrap it because we already handle `TypeError` in `isInputError`.

src/common/error.ts
```diff
@@ -0,0 +1,64 @@
...
+        return false;
+    }
+    // https://nodejs.org/api/errors.html#common-system-errors
+    // ENOTFOUND: DNS lookup failed, ENOENT: no such file or directory, ECONNREFUSED: connection refused, ECONNRESET: connection reset by peer, ETIMEDOUT: connection timed out
+    if (error.code === "ENOTFOUND" || error.code === "ENOENT" || error.code === "ECONNREFUSED" || error.code === "ECONNRESET" || error.code === "ETIMEDOUT") {
```
avanigupta, line 42: Do we need to check if `error.code !== undefined` first?

src/common/error.ts
```diff
@@ -0,0 +1,64 @@
...
+/**
+ * Check if the error is an instance of ArgumentError, TypeError, or RangeError.
+ */
+export function isInputError(error: any): boolean {
+    if (error instanceof ArgumentError ||
+        error instanceof TypeError ||
+        error instanceof RangeError) {
+        return true;
+    }
+    return false;
```
avanigupta, line 63:
```suggestion
    return (error instanceof ArgumentError ||
            error instanceof TypeError ||
            error instanceof RangeError);
```

src/keyvault/AzureKeyVaultKeyValueAdapter.ts
zhiyuanliang-ms: > Are you asking about unauthorized error from key vault? None of the errors from KeyVault are failoverable. The reason is that failing over to a different AppConfig replica will not help with any KeyVault error.

I mean the unauthorized error from `ConfigurationClient` from app config sdk.

src/common/error.ts
zhiyuanliang-ms: AuthenticationError and general error type.

src/common/error.ts
zhiyuanliang-ms: Added.

zhiyuanliang-ms:
Will send out another pr to make all error messages const in a seperated file.

src/keyvault/AzureKeyVaultKeyValueAdapter.ts
avanigupta: .net sdk always throws RequestFailedException for unauthorized error. This is a failoverable error as specified here: https://github.com/Azure/AppConfiguration-DotnetProvider/blob/bf8b06b9dde1bb219b625939b3d2ea00bd2a61d5/src/Microsoft.Extensions.Configuration.AzureAppConfiguration/AzureAppConfigurationProvider.cs#L1237

src/common/error.ts
avanigupta: In .NET, [UnauthorizedAccessException](https://learn.microsoft.com/en-us/dotnet/api/system.unauthorizedaccessexception?view=net-9.0) is not the same as 401 error from AppConfig service. Any error from AppConfig service will be thrown as `RequestFailedException` with a status code indicating what the error was (eg. 401 error code for unauthorized appconfig access).

avanigupta: It sounds like there is a difference between JS and .NET SDK error types. Any service error is thrown as `RequestFailedException` in .NET. But in JS, auth related service error (401/403) will be thrown as `AuthenticationError` and all other service errors will be thrown as `RestError`. Is my understanding correct?

src/AzureAppConfigurationImpl.ts
```diff
@@ -225,13 +230,40 @@ export class AzureAppConfigurationImpl implements AzureAppConfiguration {
...
+            if (!isInputError(error)) {
+                const timeElapsed = Date.now() - startTimestamp;
+                if (timeElapsed < MIN_DELAY_FOR_UNHANDLED_FAILURE) {
+                    // load() method is called in the application's startup code path.
+                    // Unhandled exceptions cause application crash which can result in crash loops as orchestrators attempt to restart the application.
+                    // Knowing the intended usage of the provider in startup code path, we mitigate back-to-back crash loops from overloading the server with requests by waiting a minimum time to propagate fatal errors.
+                    await new Promise(resolve => setTimeout(resolve, MIN_DELAY_FOR_UNHANDLED_FAILURE - timeElapsed));
+                }
+            }
+            throw new Error(`Failed to load: ${error.message}`);
```
avanigupta, line 263: Can we preserve stack trace before throwing?

avanigupta (approved):

src/AzureAppConfigurationImpl.ts
zhiyuanliang-ms: yes, updated.

src/common/error.ts
zhiyuanliang-ms: > But in JS, auth related service error (401/403) will be thrown as AuthenticationError and all other service errors will be thrown as RestError.

No, `AuthenticationError` is from `@azure/identity` rather than `@azure/core-http-pipeline`. This is the same in .NET.
image
A w'rong token credential will cause `AuthenticationError`. `AuthenticationError` is not the same as RestError with status 401.

If I am using a connection string with a wrong secret, I will get RestError with 401:
image

avanigupta: Moving this discussion offline since its not blocking the current PR.

avanigupta (approved):

PR #170, opened 2025-02-18
Silently fail when srv query fail

zhiyuanliang-ms:
## Why this PR?

- Fix the potential memory leak issue introduced by `Promise.race`.

- If srv query failed or timed out, it should not fail the whole provider, instead, it should be a silent error.

### DIFF
--- a/src/ConfigurationClientManager.ts
+++ b/src/ConfigurationClientManager.ts
@@ -118,8 +118,7 @@ export class ConfigurationClientManager {
             (!this.#dynamicClients ||
             // All dynamic clients are in backoff means no client is available
             this.#dynamicClients.every(client => currentTime < client.backoffEndTime) ||
-            currentTime >= this.#lastFallbackClientRefreshTime + FALLBACK_CLIENT_REFRESH_EXPIRE_INTERVAL)
-        ) {
+            currentTime >= this.#lastFallbackClientRefreshTime + FALLBACK_CLIENT_REFRESH_EXPIRE_INTERVAL)) {
             this.#lastFallbackClientRefreshAttempt = currentTime;
             await this.#discoverFallbackClients(this.endpoint.hostname);
             return availableClients.concat(this.#dynamicClients);
--- a/src/ConfigurationClientManager.ts
+++ b/src/ConfigurationClientManager.ts
@@ -18,7 +18,9 @@ const SECRET_KEY_NAME = "Secret";
 const TRUSTED_DOMAIN_LABELS = [".azconfig.", ".appconfig."];
 const FALLBACK_CLIENT_REFRESH_EXPIRE_INTERVAL = 60 * 60 * 1000; // 1 hour in milliseconds
 const MINIMAL_CLIENT_REFRESH_INTERVAL = 30 * 1000; // 30 seconds in milliseconds
-const SRV_QUERY_TIMEOUT = 30 * 1000; // 30 seconds in milliseconds
+const DNS_RESOLVER_TIMEOUT = 1_000; // 1 second in milliseconds
+const DNS_RESOLVER_TRIES = 2;
+const MAX_RESOLVESRV_TIMES = 15;
 
 export class ConfigurationClientManager {
     #isFailoverable: boolean;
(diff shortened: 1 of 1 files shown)


src/ConfigurationClientManager.ts
```diff
@@ -18,7 +18,9 @@ const SECRET_KEY_NAME = "Secret";
 const TRUSTED_DOMAIN_LABELS = [".azconfig.", ".appconfig."];
 const FALLBACK_CLIENT_REFRESH_EXPIRE_INTERVAL = 60 * 60 * 1000; // 1 hour in milliseconds
 const MINIMAL_CLIENT_REFRESH_INTERVAL = 30 * 1000; // 30 seconds in milliseconds
-const SRV_QUERY_TIMEOUT = 30 * 1000; // 30 seconds in milliseconds
+const DNS_RESOLVER_TIMEOUT = 1_000; // 1 second in milliseconds
+const DNS_RESOLVER_TRIES = 2;
+const MAX_RESOLVESRV_TIMES = 15;
```
RichardChen820, line 23: MAX_ALTNATIVE_SRV_COUNT

src/ConfigurationClientManager.ts
```diff
@@ -142,20 +145,16 @@ export class ConfigurationClientManager {
...
-                new Promise((_, reject) => timeout = setTimeout(() => reject(new Error("SRV record query timed out.")), SRV_QUERY_TIMEOUT)),
-                this.#querySrvTargetHost(host)
-            ]);
+            result = await this.#querySrvTargetHost(host);
         } catch (error) {
-            throw new Error(`Failed to build fallback clients, ${error.message}`);
-        } finally {
-            clearTimeout(timeout);
+            console.warn(`Failed to build fallback clients, ${error.message}`);
+            this.#lastFallbackClientRefreshTime = Date.now();
```
RichardChen820, line 153: Since it fails to refresh the fallback client, should not update the `lastFallbackClientRefreshTime`

src/ConfigurationClientManager.ts
```diff
@@ -18,7 +18,9 @@ const SECRET_KEY_NAME = "Secret";
 const TRUSTED_DOMAIN_LABELS = [".azconfig.", ".appconfig."];
 const FALLBACK_CLIENT_REFRESH_EXPIRE_INTERVAL = 60 * 60 * 1000; // 1 hour in milliseconds
 const MINIMAL_CLIENT_REFRESH_INTERVAL = 30 * 1000; // 30 seconds in milliseconds
-const SRV_QUERY_TIMEOUT = 30 * 1000; // 30 seconds in milliseconds
+const DNS_RESOLVER_TIMEOUT = 1_000; // 1 second in milliseconds
```
RichardChen820, line 21: What's the consideration changing it to 1 second?

src/ConfigurationClientManager.ts
```diff
@@ -182,8 +181,9 @@ export class ConfigurationClientManager {
         const results: string[] = [];
 
         try {
+            const resolver = new this.#dns.Resolver({timeout: DNS_RESOLVER_TIMEOUT, tries: DNS_RESOLVER_TRIES});
```
linglingye001, line 184: If fails at first time, will it throw error? And will the retry only happens in timeout scenario?

src/ConfigurationClientManager.ts
zhiyuanliang-ms: We should have a timeout mechanism. This 1 sec timeout is for a single resolve srv operation. The previous 30 secs timeout is for the whole srv query operation (which previously had a while(true) loop there, it can be a potential memory leak as the promise will not be canceled by Promise.race)

What amout of time will you suggest for a single resolve srv operation?

src/ConfigurationClientManager.ts
zhiyuanliang-ms: We catch the error.

We are now using the retry mechanism offered by the dns module. For a single resolveSrv operation, it will retry for the times specified by DNS_RESOLVER_TRIES

src/ConfigurationClientManager.ts
RichardChen820: 1 second is probably too short, 5 seconds make more sense to me.

src/ConfigurationClientManager.ts
zhiyuanliang-ms: I got it wrong with lastFallbackClientRefreshAttempt

src/ConfigurationClientManager.ts
zhiyuanliang-ms: Let's use 3 seconds. The worst case will be 3 * 2 retries * 10 alternatives = 1min

linglingye001 (approved):

PR #175, opened 2025-02-24
Key vault secret refresh

zhiyuanliang-ms:
## Why this PR?

Allow periodic reload of key vault secrets. In this case, even if there is no change on App Configuration key-values. This feature is targeted on the scenario where user is using latest version of a secret. In this case, the key vault reference url will not change, but the secret value will change if user rotates their secret.

reference: [#249](https://github.com/Azure/AppConfiguration-DotnetProvider/pull/249)


src/refresh/RefreshTimer.ts
```diff
@@ -3,22 +3,22 @@
 
 export class RefreshTimer {
     #backoffEnd: number; // Timestamp
-    #interval: number;
+    readonly interval: number;
```
CsCherrYY, line 6: Why not keep it private and use a get function to access it?

src/AzureAppConfigurationImpl.ts
```diff
@@ -332,8 +339,8 @@ export class AzureAppConfigurationImpl implements AzureAppConfiguration {
      * Registers a callback function to be called when the configuration is refreshed.
      */
     onRefresh(listener: () => any, thisArg?: any): Disposable {
-        if (!this.#refreshEnabled && !this.#featureFlagRefreshEnabled) {
-            throw new InvalidOperationError("Refresh is not enabled for key-values or feature flags.");
+        if (!this.#refreshEnabled && !this.#featureFlagRefreshEnabled && !this.#secretRefreshEnabled) {
+            throw new InvalidOperationError("Refresh is not enabled for key-values, key vault secrets or feature flags.");
```
CsCherrYY, line 343: I'm confused. The exception message is `key-values, key vault secrets or feature flags`, but the condition is using `&&`, so only all the three refreshes are not enabled, we will throw this exception?

src/refresh/RefreshTimer.ts
zhiyuanliang-ms: > use a get function to access it
This is anti JS pattern. Getter pattern is not common in js world.

src/AzureAppConfigurationImpl.ts
zhiyuanliang-ms: > so only all the three refreshes are not enabled, we will throw this exception?

Yes

> Refresh is not enabled for key-values, key vault secrets or feature flags.

I think “and" should be replaced with "or" in negative sentence.

src/keyvault/KeyVaultOptions.ts
```diff
@@ -4,6 +4,8 @@
 import { TokenCredential } from "@azure/identity";
 import { SecretClient } from "@azure/keyvault-secrets";
 
+export const MIN_SECRET_REFRESH_INTERVAL_IN_MS = 60_000;
```
linglingye001, line 7: Can you confirm what's the minimal secret refresh interval? 1 second or 1 minute

zhiyuanliang-ms: https://github.com/Azure/AppConfiguration-DotnetProvider/blob/main/src/Microsoft.Extensions.Configuration.AzureAppConfiguration/Constants/RefreshConstants.cs#L19

src/AzureAppConfigurationImpl.ts
```diff
@@ -483,19 +493,23 @@ export class AzureAppConfigurationImpl implements AzureAppConfiguration {
...
 
         if (this.#requestTracingEnabled && this.#aiConfigurationTracing !== undefined) {
-            // Reset old AI configuration tracing in order to track the information present in the current response from server.
+            // reset old AI configuration tracing in order to track the information present in the current response from server
             this.#aiConfigurationTracing.reset();
         }
 
         // adapt configuration settings to key-values
         for (const setting of loadedSettings) {
+            if (this.#secretRefreshEnabled && isSecretReference(setting)) {
```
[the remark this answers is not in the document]
zhiyuanliang-ms: The provider will stored all the raw secret references after each load/refresh of key values. When a secret refresh is triggered,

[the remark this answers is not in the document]
zhiyuanliang-ms: Updated. Will always cache secret with specific version.

src/AzureAppConfigurationImpl.ts
```diff
@@ -332,8 +339,8 @@ export class AzureAppConfigurationImpl implements AzureAppConfiguration {
      * Registers a callback function to be called when the configuration is refreshed.
      */
     onRefresh(listener: () => any, thisArg?: any): Disposable {
-        if (!this.#refreshEnabled && !this.#featureFlagRefreshEnabled) {
-            throw new InvalidOperationError("Refresh is not enabled for key-values or feature flags.");
+        if (!this.#refreshEnabled && !this.#featureFlagRefreshEnabled && !this.#secretRefreshEnabled) {
+            throw new InvalidOperationError("Refresh is not enabled for key-values, feature flags or Key Vault secrets.");
```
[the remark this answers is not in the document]
zhiyuanliang-ms: "Key Vault secret" is more appropriate as even if there is no change on the reference, we will still refresh the secret if the version is latest.

src/keyvault/AzureKeyVaultSecretProvider.ts
```diff
@@ -0,0 +1,94 @@
...
+                return cachedValue;
+            }
+            // not found in cache, get the secret value from key vault
+            const secretValue = await this.#getSecretValueFromKeyVault(secretIdentifier);
+            this.#cachedSecretValue.set(identifierKey, secretValue);
+            return secretValue;
+        }
+
+        // Always reload the secret value from key vault when the refresh timer expires.
+        const secretValue = await this.#getSecretValueFromKeyVault(secretIdentifier);
```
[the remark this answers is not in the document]
zhiyuanliang-ms: Change to 

```ts
// The map key is a combination of sourceId and version: "{sourceId}\n{version}".
const identifierKey = `${secretIdentifier.sourceId}\n${secretIdentifier.version ?? ""}`;

// If the secret has a version, always use the cached value if available.
if (secretIdentifier.version && this.#cachedSecretValue.has(identifierKey)) {
    return this.#cachedSecretValue.get(identifierKey);
}

if (this.#refreshTimer && !this.#refreshTimer.canRefresh()) {
    // If the refresh interval is not expired, return the cached value if available.
    if (this.#cachedSecretValue.has(identifierKey)) {
        return this.#cachedSecretValue.get(identifierKey);
    }
}

// Fallback to fetching the secret value from Key Vault.
const secretValue = await this.#getSecretValueFromKeyVault(secretIdentifier);
this.#cachedSecretValue.set(identifierKey, secretValue);
return secretValue;
```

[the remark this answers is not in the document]
RichardChen820: The new code looks good to me

RichardChen820 (approved):

src/keyvault/AzureKeyVaultSecretProvider.ts
```diff
@@ -0,0 +1,95 @@
...
+
+        if (this.#refreshTimer && !this.#refreshTimer.canRefresh()) {
+            // If the refresh interval is not expired, return the cached value if available.
+            if (this.#cachedSecretValue.has(identifierKey)) {
+                return this.#cachedSecretValue.get(identifierKey);
+            }
+        }
+
+        // Fallback to fetching the secret value from Key Vault.
+        const secretValue = await this.#getSecretValueFromKeyVault(secretIdentifier);
```
linglingye001, line 50: When secret refresh not configured, the provider can return cached value before fetching the unversioned secret value from Key Vault.

src/keyvault/AzureKeyVaultSecretProvider.ts
```diff
@@ -0,0 +1,95 @@
...
+    #secretClients: Map<string, SecretClient>; // map key vault hostname to corresponding secret client
+    #cachedSecretValue: Map<string, any> = new Map<string, any>(); // map secret identifier to secret value
+
+    constructor(keyVaultOptions: KeyVaultOptions | undefined, refreshTimer?: RefreshTimer) {
+        if (keyVaultOptions?.secretRefreshIntervalInMs !== undefined) {
+            if (refreshTimer === undefined) {
+                throw new ArgumentError("Refresh timer must be specified when Key Vault secret refresh is enabled.");
+            }
+            if (refreshTimer.interval !== keyVaultOptions.secretRefreshIntervalInMs) {
+                throw new ArgumentError("Refresh timer does not match the secret refresh interval.");
```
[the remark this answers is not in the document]
zhiyuanliang-ms: I want to prevent from misusing key-value refresher. From the API design perspective, secret provider is fully decoupled from the configuration provider.

src/keyvault/AzureKeyVaultSecretProvider.ts
zhiyuanliang-ms: Initial load scenario: Always fetch from Key Vault
All secret values are cached by SecretProvider

Refresh scenario:
1. secret refresh is configured: 
Only when refresh intervel is not expired or the secret has version, we serve secret value from cache
2. secret refresh is not configured
Only when secret has version, we serve secret value from cache

linglingye001: Right, the provider fetch all the secrets when initialize. For refresh scenario, reloading secrets from key vault only happens when refresh interval expired. So when time is not expired or user not configured secret refresh, the provider can just return secret value(with version or not) from cache. You can refer to https://learn.microsoft.com/en-us/azure/azure-app-configuration/reload-key-vault-secrets-dotnet#update-code-to-reload-key-vault-secrets-and-certificates.

zhiyuanliang-ms: If user wants to call the `refresh` method, they must configure regular refresh interval or secret refresh interval.
This is the current `SecretProvider.getSecret` logic
```ts
// The map key is a combination of sourceId and version: "{sourceId}\n{version}".
const identifierKey = `${secretIdentifier.sourceId}\n${secretIdentifier.version ?? ""}`;

// If the secret has a version, always use the cached value if available.
if (secretIdentifier.version && this.#cachedSecretValue.has(identifierKey)) {
    return this.#cachedSecretValue.get(identifierKey);
}

if (this.#refreshTimer && !this.#refreshTimer.canRefresh()) {
    // If the refresh interval is not expired, return the cached value if available.
    if (this.#cachedSecretValue.has(identifierKey)) {
        return this.#cachedSecretValue.get(identifierKey);
    }
}

// Fallback to fetching the secret value from Key Vault.
const secretValue = await this.#getSecretValueFromKeyVault(secretIdentifier);
this.#cachedSecretValue.set(identifierKey, secretValue);
return secretValue;
```
The above logic is:
1. if secret has version and cache is available, return the cached value.
2. If refresh interval is not expired and cache is available, return the cached value
3. Otherwise (cache is not available or refresh interval expired and key doesn't have version), reload secret from Key Vault

Case 1:
Secret refresh interval is configured and key-value refresh interval is not configured.
image
image

Only `refreshSecret` will be called. It will re-trigger `processKeyValue` for all cached raw secret reference configuration settings if secret refresh interval is expired. It will eventually call `SecretProvider.getSecret` method and it will fallback to fetching secret from Key Vault (which is expected as secret refresh interval is expired).

Case 2:
Secret refresh interval is not configured and normal key-value refresh is configured. Only `refreshKeyValues` will be called. For the `refreshKeyValues` task, when the refresh interval is expired, if any monitored keyvalue changed, a "refresh/reload all" will be triggered. But when processing secret reference configuration settings, it will still call `secretProvider.getSecret`. We will serve cached secret value for secret with version but always reload secret without version. This is expected behavior. Otherwise, how can we guranteed the secret is the latest?

Case 3:
Secret refresh interval is configured and normal key-value refresh is configured. `refreshKeyValues` and `refreshSecrets` will be called. For the `refreshKeyValues` task, when the refresh interval is expired, if any monitored keyvalue changed, a "refresh/reload all" will be triggered. But when processing secret reference configuration settings, it will still call `secretProvider.getSecret`. The logic is the same as case 1. For the `refreshSecrets` task, the logic is also the same as case 1.

zhiyuanliang-ms: > So when time is not expired or user not configured secret refresh, the provider can just return secret value(with version or not) from cache.

If the user didn't configure secret refresh, when refresh interval is not expired, calling `refresh` is no-op. If refresh interval is expired, if there is any change on the monitored key-values, a "refresh all` will be triggered, all secret reference will be reloaded from App Config and they will be resolved again. We can serve cached value for those secret with version. But for secret with no version, we should reload to make sure that we are serving the latest version.

If the user configured secret refresh, when refresh interval is not expired, calling `refresh` is no-op. If the refresh interval is expired, which part of `SecretProvider.getSecret` do you think is wrong?

src/AzureAppConfigurationImpl.ts
```diff
@@ -615,6 +632,21 @@ export class AzureAppConfigurationImpl implements AzureAppConfiguration {
         return Promise.resolve(needRefresh);
     }
 
+    async #refreshSecrets(): Promise<boolean> {
+        // if still within refresh interval/backoff, return
+        if (!this.#secretRefreshTimer.canRefresh()) {
+            return Promise.resolve(false);
+        }
+
+        for (const setting of this.#secretReferences) {
```
zhiyuanliang-ms, line 641: bug: `secretReferences` may be cleared in the `refreshKeyValues` task. Need a lock mechanism to avoid this.

src/AzureAppConfigurationImpl.ts
```diff
@@ -483,19 +493,23 @@ export class AzureAppConfigurationImpl implements AzureAppConfiguration {
      * Loads selected key-values and watched settings (sentinels) for refresh from App Configuration to the local configuration.
      */
     async #loadSelectedAndWatchedKeyValues() {
+        this.#secretReferences = []; // clear all cached key vault reference configuration settings
```
zhiyuanliang-ms, line 496: this could break `refreshSecrets` task

src/keyvault/AzureKeyVaultSecretProvider.ts
linglingye001: I see, there's secretRefreshTimer.canRefresh outside getSecret. In case 3：re-loading secrets is included in `refreshKeyValues` task, when `keyValueRefreshTimer` and `secretRefreshTimer` canRefresh at the same time, getSecret will be called in both tasks. Is it necessary to re-load the secrets without version twice?

zhiyuanliang-ms: > Is it necessary to re-load the secrets without version twice?

It will make the code path clear and simple. But to fix the bug I mentioned in other comment, I am thinking make sure when key-value refresh is trigger, refresh secret will not be triggered.

 BTW, we are not very care about efficiency of the refresh operation, as it should be called without await.

src/IKeyValueAdapter.ts
```diff
@@ -13,4 +13,9 @@ export interface IKeyValueAdapter {
      * This method process the original configuration setting, and returns processed key and value in an array.
      */
     processKeyValue(setting: ConfigurationSetting): Promise<[string, unknown]>;
+
+    /**
+     * This method is called when a change is detected in the configuration setting.
+     */
+    onChangeDetected(setting?: ConfigurationSetting): Promise<void>;
```
avanigupta, line 20: Does this method need ConfigurationSetting input? We dont use it currently.

src/keyvault/AzureKeyVaultSecretProvider.ts
```diff
@@ -0,0 +1,95 @@
...
+        this.#secretClients = new Map();
+        for (const client of this.#keyVaultOptions?.secretClients ?? []) {
+            const clientUrl = new URL(client.vaultUrl);
+            this.#secretClients.set(clientUrl.host, client);
+        }
+    }
+
+    async getSecretValue(secretIdentifier: KeyVaultSecretIdentifier): Promise<unknown> {
+        // The map key is a combination of sourceId and version: "{sourceId}\n{version}".
+        const identifierKey = `${secretIdentifier.sourceId}\n${secretIdentifier.version ?? ""}`;
```
avanigupta, line 35: Why not cache with the secret reference URL?

src/keyvault/AzureKeyVaultSecretProvider.ts
```diff
@@ -0,0 +1,95 @@
...
+        const secretValue = await this.#getSecretValueFromKeyVault(secretIdentifier);
+        this.#cachedSecretValue.set(identifierKey, secretValue);
+        return secretValue;
+    }
+
+    clearCache(): void {
+        // If the secret identifier has specified a version, it is not removed from the cache.
+        // If the secret identifier has not specified a version, it means that the latest version should be used. Remove the cached value to force a reload.
+        for (const key of this.#cachedSecretValue.keys()) {
+            if (key.endsWith("\n")) {
```
avanigupta, line 59: So versioned secrets are never cleared from cache even though they might not be referenced in the config store anymore? I think `clearCache` should delete everything from cache. Its okay to reload versioned secrets because otherwise we are creating an ever-growing cache of versioned secrets which might not even exist in KeyVault anymore.

src/IKeyValueAdapter.ts
zhiyuanliang-ms: > We dont use it currently.

Yes. I refered to .net provider's [implementation](https://github.com/Azure/AppConfiguration-DotnetProvider/blob/main/src/Microsoft.Extensions.Configuration.AzureAppConfiguration/IKeyValueAdapter.cs#L18). I can remove it, since we don't use it currently.

src/keyvault/AzureKeyVaultSecretProvider.ts
zhiyuanliang-ms: > So versioned secrets are never cleared from cache even though they might not be referenced in the config store anymore?

True.

src/keyvault/AzureKeyVaultSecretProvider.ts
avanigupta: I meant this should be the cache key: `https://<keyvault-name>.vault.azure.net/secrets/<secret-name>/<unique-version-id>`
And it looks like `sourceId` property is already the full secret url. Appending version is not required.
https://learn.microsoft.com/en-us/javascript/api/@azure/keyvault-secrets/keyvaultsecretidentifier?view=azure-node-latest

zhiyuanliang-ms: Oh, I didn't notice that. Make sense. Updated to use url as cachekey

src/keyvault/AzureKeyVaultSecretProvider.ts
```diff
@@ -0,0 +1,88 @@
...
+        if (secretIdentifier.version && this.#cachedSecretValue.has(identifierKey)) {
+            return this.#cachedSecretValue.get(identifierKey);
+        }
+
+        if (this.#secretRefreshTimer && !this.#secretRefreshTimer.canRefresh()) {
+            // If the refresh interval is not expired, return the cached value if available.
+            if (this.#cachedSecretValue.has(identifierKey)) {
+                return this.#cachedSecretValue.get(identifierKey);
+            }
+        }
```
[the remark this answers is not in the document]
zhiyuanliang-ms: Updated

PR #185, opened 2025-04-17
AI Configuration request tracing

zhiyuanliang-ms:


package.json
```diff
@@ -25,7 +25,7 @@
     "dev": "rollup --config --watch",
     "lint": "eslint src/ test/",
     "fix-lint": "eslint src/ test/ --fix",
-    "test": "mocha out/test/*.test.{js,cjs,mjs} --parallel"
+    "test": "mocha out/test/requestTracing.test.{js,cjs,mjs} --parallel"
```
CsCherrYY, line 28: Is skipping other tests by design?

zhiyuanliang-ms: good catch! That's a mistake.

src/requestTracing/utils.ts
```diff
@@ -125,9 +130,22 @@ export function createCorrelationContextHeader(requestTracingOptions: RequestTra
         keyValues.set(FM_VERSION_KEY, requestTracingOptions.fmVersion);
     }
 
-    // Compact tags: Features=LB+...
-    if (appConfigOptions?.loadBalancingEnabled) {
-        keyValues.set(FEATURES_KEY, LOAD_BALANCE_CONFIGURED_TAG);
+    // Compact tags: Features=LB+AI+AICC...
+    if (appConfigOptions?.loadBalancingEnabled || requestTracingOptions.aiConfigurationTracing?.usesAnyTracingFeature()) {
```
linglingye001, line 134: line134 is not needed since there's line 146

zhiyuanliang-ms: The similar pattern exists in .net provider: https://github.com/Azure/AppConfiguration-DotnetProvider/blob/main/src/Microsoft.Extensions.Configuration.AzureAppConfiguration/TracingUtils.cs#L184

My original intention is to make the code path easy to read. But now it turns out that the code looks like I am repeating myself. Decided to create a new function to wrap the logic (exactly the same as .net provider) and make it more easier to read.

src/AzureAppConfigurationImpl.ts
```diff
@@ -633,12 +648,33 @@ export class AzureAppConfigurationImpl implements AzureAppConfiguration {
...
         const [key, value] = await this.#processAdapters(setting);
         const trimmedKey = this.#keyWithPrefixesTrimmed(key);
         return [trimmedKey, value];
     }
 
+    #setAIConfigurationTracing(setting: ConfigurationSetting<string>): void {
+        if (this.#requestTracingEnabled && this.#aiConfigurationTracing !== undefined) {
+            const contentType = parseContentType(setting.contentType);
+            // content type: application/json; profile="https://azconfig.io/mime-profiles/ai"
+            if (isJsonContentType(contentType)) {
```
amerjusupovic, line 663: based on discussion offline, we might want to exclude the other known json content types (feature flag/key vault ref)

zhiyuanliang-ms: Your merged [PR](https://github.com/Azure/AppConfiguration-DotnetProvider/pull/646/files) didn't include that change, right?

amerjusupovic: It did but the logic was incorrect in how I organized it since I excluded the flag/kvref types inside the IsJson method. I separated this in the follow up PR, so that's more or less how I would expect it to look here too.

amerjusupovic (approved):

PR #188, opened 2025-04-23
Support tag filter

zhiyuanliang-ms:
## Why this PR?

Extends the existing `SettingSelector` type for `AzureAppConfigurationOptions.selectors` and `FeatureFlagOptions.selectors`.

```ts
export type SettingSelector = {
    keyFilter?: string,
    labelFilter?: string
    // New Field
    tagFilters?: string[]
    snapshotName?: string
};
```

Now, you can pass tag filters that will be used to filter key-values and feature flags. Key-values and feature flags will only be returned if they contain all of the tags and corresponding values passed to the `tagFilters` parameter.

## Usage

```js
    selectors: [
        { 
            keyFilter: "*",
            tagFilters: ["tag1=someValue1", "tag2=someValue2"] 
        }
    ],
    featureFlagOptions: {
        enabled: true,
        selectors: [
            { 
                keyFilter: "*",
                tagFilters: ["tag1=someValue1", "tag2=someValue"] 
            }
        ]
    }
});
```
### Filtering for tags with empty/null value

You can construct a tag filter string like `"EmptyTag="` and `"NullTag=\0"`.

```js
    selectors: [
        { 
            keyFilter: "*",
            tagFilters: ["EmptyTag=", "NullTag=\0"] 
        }
    ]
});
```


src/types.ts
```diff
@@ -59,3 +69,13 @@ export enum LabelFilter {
...
 }
+
+/**
+ * TagFilter is used to filter key-values based on tags.
+ */
+export enum TagFilter {
+    /**
+     * Matches key-values without a label.
+     */
+    Null = ""
```
RichardChen820, line 80: Is this better?
```
export enum TagValue {
    /**
     * Represent empty tag value
     */
    Empty = ""
}
```

src/AzureAppConfigurationImpl.ts
```diff
@@ -899,11 +907,11 @@ function getValidSettingSelectors(selectors: SettingSelector[]): SettingSelector
...
         if (selector.snapshotName) {
-            if (selector.keyFilter || selector.labelFilter) {
-                throw new ArgumentError("Key or label filter should not be used for a snapshot.");
+            if (selector.keyFilter || selector.labelFilter || selector.tagFilters) {
+                throw new ArgumentError("Key, label or tag filter should not be used for a snapshot.");
             }
         } else {
-            if (!selector.keyFilter) {
+            if (!selector.keyFilter && (!selector.tagFilters || selector.tagFilters.length === 0)) {
                 throw new ArgumentError("Key filter cannot be null or empty.");
```
RichardChen820, line 915: Key filter and tag filter cannot both be null or empty.

src/AzureAppConfigurationImpl.ts
```diff
@@ -912,11 +920,31 @@ function getValidSettingSelectors(selectors: SettingSelector[]): SettingSelector
...
+        return false;
+    }
+    if (tagsA.length !== tagsB.length) {
+        return false;
+    }
+
+    const sortedStringA = [...tagsA].sort().join("\n");
+    const sortedStringB = [...tagsB].sort().join("\n");
+
+    return sortedStringA === sortedStringB;
```
RichardChen820, line 945: Is this check robust? `\n` is not allowed in tag?

src/AzureAppConfigurationImpl.ts
```diff
@@ -58,6 +58,8 @@ import { InvalidOperationError, ArgumentError, isFailoverableError, isInputError
 
 const MIN_DELAY_FOR_UNHANDLED_FAILURE = 5_000; // 5 seconds
 
+const MAX_TAG_FILTERS = 5;
```
RichardChen820, line 61: MAX_TAG_FILTERS_COUNT

src/types.ts
zhiyuanliang-ms: I copied the comment from label filter. I forgot to update it. Definitly better.

src/AzureAppConfigurationImpl.ts
zhiyuanliang-ms: I followed .NET provider's implementation: https://github.com/Azure/AppConfiguration-DotnetProvider/pull/637/files#diff-5f9d810c153f73ab4382bfddf40b76b7c30bba39beae6e74557bc078115f3f95

I checked out backend [code](https://msazure.visualstudio.com/Azure%20AppConfig/_search?action=contents&text=ParseTagFilter&type=code&lp=code-Project&filters=ProjectFilters%7BAzure%20AppConfig%7DRepositoryFilters%7BAppConfigService%7D&pageSize=25&result=DefaultCollection/Azure%20AppConfig/AppConfigService/GBdev//src/Microsoft.AppConfig.Service/Utils/SearchQueryHelper.cs), we don't forbid `\n` but the portal will not allow you to enter `\n` when creating a tag.

src/AzureAppConfigurationImpl.ts
```diff
@@ -921,24 +929,44 @@ function getValidSettingSelectors(selectors: SettingSelector[]): SettingSelector
     return uniqueSelectors.map(selectorCandidate => {
         const selector = { ...selectorCandidate };
         if (selector.snapshotName) {
-            if (selector.keyFilter || selector.labelFilter) {
-                throw new ArgumentError("Key or label filter should not be used for a snapshot.");
+            if (selector.keyFilter || selector.labelFilter || selector.tagFilters) {
+                throw new ArgumentError("Key, label or tag filter should not be used for a snapshot.");
```
avanigupta, line 933:
```suggestion
                throw new ArgumentError("Key, label or tag filters should not be specified while selecting a snapshot.");
```

src/AzureAppConfigurationImpl.ts
```diff
@@ -959,3 +987,15 @@ function getValidFeatureFlagSelectors(selectors?: SettingSelector[]): SettingSel
     });
     return getValidSettingSelectors(selectors);
 }
+
+function validateTagFilters(tagFilters: string[]): void {
+    if (tagFilters.length > MAX_TAG_FILTER_COUNT) {
+        throw new Error(`The number of tag filters cannot exceed ${MAX_TAG_FILTER_COUNT}.`);
+    }
```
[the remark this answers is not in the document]
zhiyuanliang-ms: Removed.

RichardChen820 (approved):

test/load.test.ts
```diff
@@ -81,12 +81,24 @@ const mockedKVs = [{
...
     tags: {"tag1": "someValue"}
 }, {
     key: "keyWithTag2",
-    value: "someValue",
+    value: "someValue2",
     tags: {"tag2": "someValue"}
+}, {
+    key: "keyWithNullTag",
+    value: "valueWithNullTag",
+    tags: {"nullTag": "\0"}
```
avanigupta, line 93: The tag value should be `null` since this is json. `"\0"` is a string value (not the same as `null`).

```
     tags: {"nullTag": null}
```

avanigupta: A key-value with null/empty tag value looks like this in AppConfig:

```json
{
	"key": "TestNullTags",
	"label": null,
	"value": "example value",
	"content_type": null,
	"tags": {
		"nullTagValue": null,
		"emptyTagValue": ""
	}
}
```

To select this specific key-value, the user must provide the following selectors:
```
    selectors: [{
          keyFilter: "Test*",
          tagFilters: ["nullTagValue=\0", "emptyTagValue="] 
       }
    ]
});
```

avanigupta:
I dont see `keyFilter` specified in the usage examples in PR description. Key filter is not optional, so please make sure that missing key filter should be throwing an error. And also add key filter to the sample code snippets.

test/load.test.ts
zhiyuanliang-ms: Updated.

zhiyuanliang-ms: BTW, match "\0" in query string with `null` in json value is done by our backend, right?

I saw the [testcase](https://github.com/Azure/AppConfiguration-DotnetProvider/blob/main/tests/Tests.AzureAppConfiguration/Unit/TagFiltersTests.cs#L207) in .net provider and I did the same thing to the _filterKVs helper method.

avanigupta (approved):

Issue #191, opened 2025-04-30
Secret loading is sequential and missing pre-emptive authentication


ronfor:
### Summary

While using the library to load App Configuration, I’ve observed slow performance due to two key issues:

- Secrets are fetched from Key Vault **sequentially**
- No **pre-emptive authentication** is performed

### Investigation

This behaviour was identified using Application Insights, with traces collected via OpenTelemetry on a Node.js process.

Below is an example trace showing a single configuration lookup referencing 14 secrets (which are stored in 4 x separate vaults):
![Image](https://github.com/user-attachments/assets/07ae5152-da00-4b1a-ba52-e6687bcc2aa4)

The `loadConfig` trace wraps the point we call the `load` function: https://github.com/Azure/AppConfiguration-JavaScriptProvider/blob/5c1f4a34475fe94326228f8bdffaa8088779a2d6/src/load.ts#L28

Each secret fetch initially results in a 401 (unauthenticated), followed by a successful retry after authentication. This happens for every unique Key Vault endpoint used.

Despite the environment running within Azure, the full configuration load takes approximately **2 seconds**, introducing noticeable latency.

### Expected behaviour

Performance could be improved significantly if:

- Secrets were fetched **in parallel**
- Authentication was performed **proactively**, avoiding repeated 401/retry cycles

zhiyuanliang-ms:
Hi, @ronfor. Thanks for reaching out!

> No pre-emptive authentication is performed

This behavior of Azure Key Vault is by designed. Key Vault backend uses a challenge-based authentication. Please see this issue [#31228](https://github.com/Azure/azure-sdk-for-python/issues/31228)

> Secrets are fetched from Key Vault sequentially

We are looking into make App Config provider fetch Key Vault secret in parallel. I will come back to you after I implement the POC.

zhiyuanliang-ms:
Hi, @ronfor 

I write a POC in this branch [zhiyuanliang/secret-performance](https://github.com/Azure/AppConfiguration-JavaScriptProvider/pull/192)

Can you try this private build and check to what extent it can improve the performance.



I added 15 key vault reference in my App Config store, 

![Image](https://github.com/user-attachments/assets/06747bad-432f-4ed6-9190-381be97332d7)

```javascript
import { DefaultAzureCredential } from "@azure/identity";
const credential = new DefaultAzureCredential();

import { load } from "@azure/app-configuration-provider";
[...]

PR #192, opened 2025-05-07
Resolve key vault secret in parallel

zhiyuanliang-ms:
## Why this PR?

#191 

This change can bring about 40% speed-up when loading 10+ secrets.

## Visible change

Added `KeyVaultOptions.parallelSecretResolutionEnabled` option to allow resolving key vault secret in parallel.

Usage:

```js
const credential = new DefaultAzureCredential();
const appConfig = await load("endpoint", credential, {
  keyVaultOptions: { credential: credential,  parallelSecretResolutionEnabled: true}
});
```

## About Key Vault throttling

Key Vault allows 4000 transactions in 10 seconds, per vault per region. [ref](https://learn.microsoft.com/en-us/azure/key-vault/general/service-limits#secrets-managed-storage-account-keys-and-vault-transactions)

1. Key Vault SDK cannot be used in browser because of CORS. I have verified this. So, for the potential browser scenario of JS provider, people should not use Key Vault reference there. This significantly reduces the chance of Key Vault throttling as Key Vault reference is only valid to use in server scenario.

2. For potential throttling response from Key Vault, the [best practice](https://learn.microsoft.com/en-us/azure/key-vault/general/overview-throttling#how-to-throttle-your-app-in-response-to-service-limits) is to use the built-in retry policy configured through SecretClientOptions.

The provider supports either registering `SecretClient` instances or configuring `SecretClientOptions`. [ref](https://github.com/Azure/AppConfiguration-JavaScriptProvider/blob/main/src/keyvault/KeyVaultOptions.ts#L27)

## About cache for Key Vault secret
Secret cache is added in #175


src/AzureAppConfigurationImpl.ts
```diff
@@ -494,11 +494,22 @@ export class AzureAppConfigurationImpl implements AzureAppConfiguration {
...
+                    keyValues.push([key, value]);
+                });
+                secretResolutionPromises.push(secretResolutionPromise);
+                continue;
+            }
             const [key, value] = await this.#processKeyValue(setting);
             keyValues.push([key, value]);
         }
+        // wait for all secret resolution promises to be resolved
+        await Promise.all(secretResolutionPromises);
```
juniwang, line 512: Thanks for the quick POC. I would be very careful to enable parallel secret resolution that makes it easier to hit KeyVault limits. We need to make sure:
- we and the sdk have proper retry policies in case of 429, honor `Retry-After` if present
- we may need to cache the secrets for certain time, combined with configurable cache expiration
- and it would be useful if we can control the parallisim to decrease the possibility of 429.

zhiyuanliang-ms: > we may need to cache the secrets for certain time, combined with configurable cache expiration

Secret cache is added in this PR: #175

zhiyuanliang-ms: Key Vault allows 4000 transactions in 10 seconds, per vault per region.
https://learn.microsoft.com/en-us/azure/key-vault/general/service-limits#secrets-managed-storage-account-keys-and-vault-transactions


1. Key Vault SDK cannot be used in browser because of CORS. I have verified this. So, for the potential browser scenario of JS provider, people should not use Key Vault reference there. This significantly reduces the chance of Key Vault throttling as Key Vault reference is only valid to use in server scenario.

2. For potential throttling response from Key Vault, we have two options:

- Key Vault SDK has built-in retry policy which can be configured through SecretClientOptions
https://learn.microsoft.com/en-us/azure/key-vault/general/overview-throttling

The configuration provider supports three ways to connect to resolve Key Vault secret

```ts
export interface KeyVaultOptions {
    /**
     * Specifies the Key Vault secret client used for resolving Key Vault references.
     */
    secretClients?: SecretClient[];

    /**
     * Specifies the credentials used to authenticate to key vaults that have no applied SecretClient.
     */
    credential?: TokenCredential;

    /**
     * Specifies the callback used to resolve key vault references that have no applied SecretClient.
     * @param keyVaultReference The Key Vault reference to resolve.
     * @returns The secret value.
     */
    secretResolver?: (keyVaultReference: URL) => string | Promise<string>;
}
```

1. user provides credential, then provider will use this credential and create a SecretClient with default SecretClientOptions
2. user provides SecretClient instance (in this case, user can configure the SecretClientOptions)
3. user provides a custom secretResolver callback function which takes the key vault reference url and return the resolved secret value

There is a missing piece in JS provider,
We allow user to set SecretClientOptions in .NET provider: https://github.com/Azure/AppConfiguration-DotnetProvider/pull/577
We can offer the same functionality for JS provider



- Implement the retry logic by ourselves, the benefit is that we can make sure we eventually can get the secret value, but it will significantly increase the code complexity

Besides, only the initial load scenario is important, but for the initial load scenario, we've already have startup retry mechanism

For the refresh scenario, even if we are throttled by the key vault, we still have the cached secret value to serve.

drago-draganov: My expectation is that we offer this behind optional setting. Customers to opt-in for parallel secrets materialization.

drago-draganov: enableParallelSecretResolution

jimmyca15: @drago-draganov that sounds like a function name. Do you want it to be a function?

drago-draganov: By no means. Use whatever we have currently have as pattern.

jimmyca15: Okay. For boolean fields we usually have enabled at the end, like [here](https://github.com/Azure/AppConfiguration-JavaScriptProvider/blob/28cbd6d975e31d43360503e5a8d445fdc00383bf/src/AzureAppConfigurationOptions.ts#L68C1-L69C1).

drago-draganov: Solid choice.

CsCherrYY (approved):

PR #214, opened 2025-08-27
Collect node.js version

zhiyuanliang-ms:


juniwang:
> @zhenlan Closed as I found that JS Azure SDK has already collected the Node version. [Code](https://github.com/Azure/azure-sdk-for-js/blob/1cd135255d4804e9164ea47fc5b23a1e17e22af1/sdk/core/core-rest-pipeline/src/util/userAgentPlatform.ts#L33)

Do you mean we already have this data in Geneva/kusto?

zhiyuanliang-ms:
@juniwang Yes

PR #216, opened 2025-08-28
Add missing semicolon

zhiyuanliang-ms:


CsCherrYY (approved):

examples/console-app/configObject.mjs
```diff
@@ -2,7 +2,7 @@
 // Licensed under the MIT license.
 
 import * as dotenv from "dotenv";
-dotenv.config()
+dotenv.config();
```
juniwang, line 5: have we enabled coding style tools and possible auto fixing?

And you are adding semicolon, not comma

zhiyuanliang-ms: We added linting check for the source code, the example folder is not included.

juniwang: do we need to manually include any folder for linting? and why `examples` is not included?

zhiyuanliang-ms: `examples` folder has its own package.json file. I will add lint check for it.

>do we need to manually include any folder for linting?

Yes. https://github.com/Azure/AppConfiguration-JavaScriptProvider/blob/main/package.json#L26

CsCherrYY (approved):

PR #232, opened 2025-10-02
Support snapshot reference

zhiyuanliang-ms:
## Snapshot Reference

A snapshot reference is a configuration setting that references a snapshot in the same App Configuration store. When loaded, the provider resolves it and adds all key-values from that snapshot. Using snapshot references enables switching between snapshots at runtime, unlike using a snapshot selector, which requires code changes and/or restarts to switch to a new snapshot.

The content type of a snapshot reference is: `"application/json; profile=\"https://azconfig.io/mime-profiles/snapshot-ref\"; charset=utf-8";`

## Behavior difference from .NET provider
Different from PR [#689](https://github.com/Azure/AppConfiguration-DotnetProvider/pull/689)

The snapshot api design in JS provider is different from .NET provider. 
In JS provider, keyvalue and feature flag are fully decoupled. The load and refresh operations for key values and feature flags are performed separatedly. The selector for key value will not load feature flags, vice versa. You must explicitly specify selectors for them.

To load key value or feature flag from snapshot, you must use a snapshot selector i.e. `SettingSelector.snapshotName`. When you specify a snapshot selector for key values, even if there is any feature flag in that snapshot, the feature flag will not be loaded when loading key values. You must also specify the same snapshot selector in `FeatureFlagOptions`.

It is a library level choice. We won't break it. So, when resolving a snapshot reference, feature flags in the snapshot will be ignored.

**Note**: This PR is not ready for review yet. It depends on the Azure JS SDK adding support for snapshot reference types. After that update, I’ll replace all existing helper functions with the SDK ones.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -3681,9 +3681,9 @@
       "license": "MIT"
     },
     "node_modules/js-yaml": {
-      "version": "4.1.0",
-      "resolved": "https://registry.npmjs.org/js-yaml/-/js-yaml-4.1.0.tgz",
-      "integrity": "sha512-wpxZs9NoxZaJESJGIZTyDEaYpl0FKSA+FB9aJiyemKhMwkxQg63h4T1KJgUGHpTqPDNRcmmYLugrRjJlBtWvRA==",
+      "version": "4.1.1",
+      "resolved": "https://registry.npmjs.org/js-yaml/-/js-yaml-4.1.1.tgz",
+      "integrity": "sha512-qQKT4zQxXl8lLwBtHMWwaTcGfFOZviOJet3Oy/xmGk2gZH677CJM9EvtfdSkgWcATZhj/55JZ0rmy3myCT5lsA==",
       "dev": true,
       "license": "MIT",
       "dependencies": {


zhiyuanliang-ms:
Should we handle the case that sentinel key is a snapshot reference?

src/appConfigurationImpl.ts
```diff
@@ -34,9 +35,16 @@ import {
     METADATA_KEY_NAME,
     ETAG_KEY_NAME,
     FEATURE_FLAG_REFERENCE_KEY_NAME,
+    ALLOCATION_ID_KEY_NAME,
```
jimmyca15: These changes seem unrelated to snapshot references. Where'd they come from?

src/appConfigurationImpl.ts
```diff
@@ -504,17 +514,26 @@ export class AzureAppConfigurationImpl implements AzureAppConfiguration {
...
+                settings = await this.#loadConfigurationSettingsFromSnapshot(selector.snapshotName);
             }
 
             for (const setting of settings) {
+                if (isSnapshotReference(setting) && !loadFeatureFlag) {
+                    this.#useSnapshotReference = true;
+
+                    // TODO: When SDK supports snapshot reference, use the helper method from SDK.
+                    const snapshotName = parseSnapshotReference(setting).value.snapshotName;
+                    const settingsFromSnapshot = await this.#loadConfigurationSettingsFromSnapshot(snapshotName);
```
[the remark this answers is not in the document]
jimmyca15: Got it. Looks like we throw in .NET as well: https://github.com/Azure/AppConfiguration-DotnetProvider/blob/34576ef70d303a029b7ea7ef860f12885fb8f35d/src/Microsoft.Extensions.Configuration.AzureAppConfiguration/SnapshotReference/SnapshotReferenceParser.cs#L57C1-L58C1

src/appConfigurationImpl.ts
```diff
@@ -504,17 +509,29 @@ export class AzureAppConfigurationImpl implements AzureAppConfiguration {
...
             }
 
             for (const setting of settings) {
+                if (isSnapshotReference(setting) && !loadFeatureFlag) {
+                    this.#useSnapshotReference = true;
+
+                    const snapshotRef: ConfigurationSetting<SnapshotReferenceValue> = parseSnapshotReference(setting);
+                    const snapshotName = snapshotRef.value.snapshotName;
+                    if (!snapshotName) {
+                        throw new SnapshotReferenceError(`Invalid format for Snapshot reference setting '${setting.key}'.`);
```
jimmyca15: We use InvalidOperationError if a snapshot has the wrong composition type. Can't we use it for this case as well?

zhiyuanliang-ms: How about throwing SnapshotReferenceError when composition type is wrong. This will make it consistent with how we throw [KeyValueReferenceError](https://github.com/Azure/AppConfiguration-JavaScriptProvider/blob/main/src/keyvault/keyVaultKeyValueAdapter.ts#L28)

BTW, `InvalidOperationError` is not a javascript built-in error type. It is a custom error type defined by us.

PR #277, opened 2026-02-10
Version bump 2.4.0

linglingye001:


zhiyuanliang-ms:
Discussed offline. This should be a minor version bump

Issue #325, opened 2026-06-16
Dedupe concurrent Key Vault secret requests when multiple references resolve to the same secret


jimmythomson:
# Dedupe concurrent Key Vault secret requests when multiple references resolve to the same secret

## Summary

When an App Configuration store contains multiple Key Vault references that point to the same secret (same `sourceId`), the provider can issue redundant `getSecret` requests to Azure Key Vault. The in-memory cache in `AzureKeyVaultSecretProvider` only stores **resolved** values, not **in-flight** requests, so concurrent resolutions of the same secret are not deduplicated.

## Current behavior

The cache lives in `src/keyvault/keyVaultSecretProvider.ts`:

```ts
async getSecretValue(secretIdentifier) {
    const identifierKey = secretIdentifier.sourceId;
    if (this.#cachedSecretValues.has(identifierKey) &&
        (!this.#secretRefreshTimer || !this.#secretRefreshTimer.canRefresh())) {
            return this.#cachedSecretValues.get(identifierKey);
    }
    const secretValue = await this.#getSecretValueFromKeyVault(secretIdentifier);
    this.#cachedSecretValues.set(identifierKey, secretValue);
    return secretValue;
}
```

Secret references are dispatched from `AppConfigurationImpl.#resolveSecretReferences`:

- **Sequential mode** (`resolveSecretsInParallel = false`, default): references are awaited one at a time, so the first call populates the cache and subsequent duplicates are served from it. ✅ One Key Vault request.
- **Parallel mode** (`resolveSecretsInParallel = true`): all `processKeyValue` promises are started before any resolve, then awaited with `Promise.all`. Because the cache is only written *after* `await #getSecretValueFromKeyVault(...)` completes, every duplicate reference that starts before the first one returns makes its own network call. ❌ N concurrent Key Vault requests for N duplicates.

The same multiplication occurs on refresh ticks: when `secretRefreshIntervalInMs` is set and `#secretRefreshTimer.canRefresh()` is true, the cache is bypassed for every reference in that round, so duplicates fan out again.

## Impact

- Unnecessary load on Azure Key Vault (latency, throttling risk against per-vault TPS limits).
- Higher cost / quota consumption for customers with many references to the same secret.
- Slower `load()` / `refresh()` when the duplicates trigger throttling or retries.
- Inconsistent behavior between sequential and parallel modes for what should be a transparent performance toggle.

## Expected behavior

[...]

PR #328, opened 2026-06-22
Optimization for concurrent Key Vault Ref resolution - deduplicate requests to the same secret

linglingye001:
The optimization idea is from https://github.com/Azure/AppConfiguration-JavaScriptProvider/issues/325

### DIFF
--- a/src/appConfigurationImpl.ts
+++ b/src/appConfigurationImpl.ts
@@ -110,7 +110,6 @@ export class AzureAppConfigurationImpl implements AzureAppConfiguration {
     #secretRefreshEnabled: boolean = false;
     #secretReferences: ConfigurationSetting[] = []; // cached key vault references
     #secretRefreshTimer: RefreshTimer | undefined = undefined;
-    #resolveSecretsInParallel: boolean = false;
 
     /**
      * Selectors of key-values obtained from @see AzureAppConfigurationOptions.selectors
@@ -202,7 +201,6 @@ export class AzureAppConfigurationImpl implements AzureAppConfiguration {
                 this.#secretRefreshEnabled = true;
                 this.#secretRefreshTimer = new RefreshTimer(secretRefreshIntervalInMs);
             }
-            this.#resolveSecretsInParallel = options.keyVaultOptions.parallelSecretResolutionEnabled ?? false;
         }
         this.#adapters.push(new AzureKeyVaultKeyValueAdapter(options?.keyVaultOptions, this.#secretRefreshTimer));
         this.#adapters.push(new JsonKeyValueAdapter());
@@ -558,23 +556,17 @@ export class AzureAppConfigurationImpl implements AzureAppConfiguration {
             this.#aiConfigurationTracing.reset();
         }
 
+        for (const adapter of this.#adapters) {
+            await adapter.preload?.(loadedSettings.filter(s => adapter.canProcess(s))); // dedup and warm the secret cache
+        }
+
         for (const setting of loadedSettings) {
             if (isSecretReference(setting)) {
                 this.#secretReferences.push(setting); // cache secret references for resolve/refresh secret separately
-                continue;
             }
             // adapt configuration settings to key-values
             const [key, value] = await this.#processKeyValue(setting);
             keyValues.push([key, value]);
-        }
-
-        if (this.#secretReferences.length > 0) {
-            for (const adapter of this.#adapters) {
-                await adapter.preload?.(this.#secretReferences); // dedup and warm the secret cache
-            }
-            await this.#resolveSecretReferences(this.#secretReferences, (key, value) => {
-                keyValues.push([key, value]);
-            });
         }
 
         this.#clearLoadedKeyValues(); // clear existing key-values in case of configuration setting deletion
@@ -674,7 +666,7 @@ export class AzureAppConfigurationImpl implements AzureAppConfiguration {
 
         if (needRefresh) {
             for (const adapter of this.#adapters) {
-                await adapter.onChangeDetected();
+                await adapter.onChangeDetected?.();
             }
             await this.#loadSelectedKeyValues();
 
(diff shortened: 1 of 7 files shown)


src/keyvault/keyVaultKeyValueAdapter.ts
```diff
@@ -49,6 +49,28 @@ export class AzureKeyVaultKeyValueAdapter implements IKeyValueAdapter {
         }
     }
 
+    /**
+     * Returns the normalized Key Vault secret identifier (sourceId) for a secret reference setting,
+     * or undefined if the reference cannot be parsed. Used to deduplicate references that resolve to
+     * the same secret before resolving them.
+     */
+    getSecretReferenceId(setting: ConfigurationSetting): string | undefined {
```
zhiyuanliang-ms: This is a stateless operation which will be used outside. It should not be a method under key vault kv adapter.

src/appConfigurationImpl.ts
```diff
@@ -715,6 +717,8 @@ export class AzureAppConfigurationImpl implements AzureAppConfiguration {
             return Promise.resolve(false);
         }
 
+        // Invalidate cached secret values so the refresh round re-fetches them from Key Vault.
+        this.#keyVaultAdapter.clearCache();
```
zhiyuanliang-ms: With the introduction of features such as secret refresh, parallel loading, and deduplication, `AppConfigurationImpl` now needs more fine-grained control over secret loading and caching.

When `KeyValueAdapter.onChangeDetected` is available, but we bypass it and directly call `SecretProvider` or `Adapter.clearCache`, I do not think the existing `KeyVaultKeyValueAdapter` abstraction remains a sound design.

zhiyuanliang-ms: I think we can introduce `preload(settings: ConfigurationSetting[]): Promise<void>;` to `IKeyVaultAdapter`

This method will dedup and warm the secret cache.

SecretProvider needs to be updated. `getSecretValue` should now only serve secret value from cache (if no cache is found, fallback to send request), the timer gate should be removed.

The timer gate should apply for the preload path only.

linglingye001: Discussed offline

src/keyValueAdapter.ts
```diff
@@ -14,6 +14,12 @@ export interface IKeyValueAdapter {
      */
     processKeyValue(setting: ConfigurationSetting): Promise<[string, unknown]>;
 
+    /**
+     * Optionally batch-resolves settings ahead of processKeyValue, e.g. to deduplicate and warm up
+     * Key Vault secret requests so that processKeyValue only reads from cache.
+     */
+    preload?(settings: ConfigurationSetting[]): Promise<void>;
```
zhiyuanliang-ms, line 20: This method should not be optional. We should make it and its comment consistent with `OnChangeDetected`.

src/keyvault/keyVaultKeyValueAdapter.ts
```diff
@@ -49,6 +49,32 @@ export class AzureKeyVaultKeyValueAdapter implements IKeyValueAdapter {
...
     }
 
+    /**
+     * Deduplicates secret references by their normalized secret identifier (sourceId) and preloads each
+     * unique secret exactly once, warming the cache so that processKeyValue only reads from it.
+     * Best-effort: unparseable references are skipped and re-surfaced by processKeyValue with full context.
+     */
+    async preload(settings: ConfigurationSetting[]): Promise<void> {
+        if (!this.#keyVaultOptions) {
+            return; // nothing to do; processKeyValue will throw the proper ArgumentError
```
zhiyuanliang-ms:
```suggestion
            return; // no-op when keyVaultOptions is not configured
```

src/keyValueAdapter.ts
zhiyuanliang-ms: In the comment we should explain the parameter: whether it should be just a bunch of configuration settings or a set of certain type of configuration settings that the adapter can process

src/keyvault/keyVaultSecretProvider.ts
```diff
@@ -33,16 +33,53 @@ export class AzureKeyVaultSecretProvider {
         }
     }
 
+    /**
+     * Fetches the given unique secrets ahead of resolution to warm the cache. Honors the secret refresh
+     * timer and the parallel resolution option. This is best-effort: per-secret failures are swallowed so
+     * that the error is surfaced with full context later by getSecretValue during resolution.
+     */
+    async preloadSecrets(secretIdentifiers: KeyVaultSecretIdentifier[]): Promise<void> {
```
zhiyuanliang-ms: preload is the concept from the abstraction of kv adapter. This business logic should be placed in adapter implementation.

I think secret provider should offer two methods: 1.  loadSecretValue 2. getSecretValue

`loadSecretValue` has refresh timer gate, if timer is not due, serve secret value from cache, when there is no cache or timer is due, the fallback is to send request to key vault to get secret value
`getSecretValue` simply serve secret value from cache, when there is no cache, the fallback is `loadSecretValue`

src/keyValueAdapter.ts
zhiyuanliang-ms: I actually prefer to let preload only take settings that the adapter can process. This will match the orchestrator(appConfigurationImpl) + kv adapter design.

The orchestrator should call preload like this:

```ts
for (const adapter of this.#adapters) {
    await adapter.preload(settings.filter(s => adapter.canProcess(s)));
}
```
The above code has very clear intention.
But the concern here is that it will increase the cost to MxN (M is adapater count, N is settings count)

zhiyuanliang-ms: With that concern, I think maybe we should make `preload` and `onChangeDetected` optional method under the `IKeyValueAdapter` to leverage the JS/TS language characteristics for better performance.

src/appConfigurationImpl.ts
```diff
@@ -569,6 +569,9 @@ export class AzureAppConfigurationImpl implements AzureAppConfiguration {
         }
 
         if (this.#secretReferences.length > 0) {
+            for (const adapter of this.#adapters) {
+                await adapter.preload?.(this.#secretReferences); // dedup and warm the secret cache
```
zhiyuanliang-ms: We should call preload hook in `loadSelectedKeyValue` like this:

```ts
for (const adapter of this.#adapters) {
   if (adapter.preload) {
	await adapter.preload(settings.filter(s => adapter.canProcess(s)));
   }
}
```

linglingye001: if `preload` is undefined, the `filter` argument is not executed. so the following code is fine
```javascript
for (const adapter of this.#adapters) {
    await adapter.preload(settings.filter(s => adapter.canProcess(s)));
}
```

src/keyValueAdapter.ts
linglingye001: discussed offline

src/keyvault/keyVaultKeyValueAdapter.ts
```diff
@@ -29,23 +29,57 @@ export class AzureKeyVaultKeyValueAdapter implements IKeyValueAdapter {
         if (!this.#keyVaultOptions) {
             throw new ArgumentError(KeyVaultReferenceErrorMessages.KEY_VAULT_OPTIONS_UNDEFINED);
         }
-        let secretIdentifier: KeyVaultSecretIdentifier;
-        try {
-            secretIdentifier = parseKeyVaultSecretIdentifier(
-                parseSecretReference(setting).value.secretId
-            );
-        } catch (error) {
-            throw new KeyVaultReferenceError(buildKeyVaultReferenceErrorMessage("Invalid Key Vault reference.", setting), { cause: error });
```
zhiyuanliang-ms, line 38: We should keep this error handling

zhiyuanliang-ms (approved):

Issue #333, opened 2026-06-26
Refresh change detection can miss updates when page._response.status is a string (`"200"`)


IvanTheBearable:
### Describe the bug
In dynamic refresh, change detection appears to rely on strict numeric comparison of `page._response.status`.

The current condition in `src/appConfigurationImpl.ts`:

```ts
if (page._response.status === 200) { // created or changed
  return true;
}
```

In my runtime, `page._response.status` is represented as a string value (`"200"`).  When this happens, refresh does not detect changes even though the service returns updated data.

### To Reproduce
1. Use `@azure/app-configuration-provider@2.5.0` with feature flag refresh enabled.
```ts
const endpoint = process.env.AZURE_APP_CONFIG_ENDPOINT;
const appConfig = await load(endpoint, new DefaultAzureCredential(), {
    selectors: [{ keyFilter: "_" }],
    featureFlagOptions: {
      enabled: true,
      selectors: [{ keyFilter: "*" }],
      refresh: {
        enabled: true,
        refreshIntervalInMs,
      },
    },
  });

  const featureProvider = new ConfigurationMapFeatureFlagProvider(appConfig);
  const featureManager = new FeatureManager(featureProvider);
```
2. Get the value of a FeatureFlag
```
await appConfig.refresh();
const isEnabled = await featureManager.isEnabled(flagName);
```
3. Change the feature flag in Azure App Configuration.
4. Get the value of the feature flag again.
5. Observe that the value does not update.

This [Gist](https://gist.github.com/IvanTheBearable/cd2c2b998302f988ec5ad98e88bc34af) contains a full sample script.

### Expected behavior
If the service response indicates changed data (HTTP 200), refresh should detect the change and reload feature flags.

### Actual behavior
Refresh requests are sent and changed data is returned, but local feature flags remain stale until process restart.

### Environment
- Package: `@azure/app-configuration-provider@2.5.0`
- App framework: Next.js App Router (Node SSR path)
- Node: `22.22.2`
- OS: macOS Darwin `25.5.0` arm64
- Related package: `@microsoft/feature-management@2.3.1`

### Candidate fix (validated locally)
Changing the comparison to coerce status resolves the issue in my environment:

```ts
if (Number(page._response.status) === 200) { // created or changed
  return true;
}
```

This local patch made `refresh()` apply feature flag updates without restart.

### Notes
I can open a PR with the above fix.

zhiyuanliang-ms:
[...]

PR #334, opened 2026-06-26
Fix feature flag refresh detection when page status is string

IvanTheBearable:
### Summary

Feature-flag refresh change detection currently assumes page response status is numeric. In some runtime paths, status is returned as a string, which prevents detecting a valid 200 response and can leave feature flags stale.

Addresses issue #333 

### What changed

- Normalize page response status with Number(...) in refresh change detection
- Add regression test support for string status responses
- Add a dynamic refresh feature-flag test that reproduces and verifies the string-status scenario

### Why

This makes change detection robust across status value types and prevents false no-change outcomes when status is "200".

### Validation

Built test outputs
Ran refresh test suite for both ESM and CJS artifacts
All tests passed, including the new regression scenario

### DIFF
--- a/src/appConfigurationImpl.ts
+++ b/src/appConfigurationImpl.ts
@@ -750,7 +750,7 @@ export class AzureAppConfigurationImpl implements AzureAppConfiguration {
 
                 for await (const page of pageIterator) {
                     // when conditional request is sent, the response will be 304 if not changed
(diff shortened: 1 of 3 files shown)


zhiyuanliang-ms:
Thanks for contribution @IvanTheBearable 
We’ve moved the fix to #335

PR #335, opened 2026-06-26
Fix refresh bug when response status is string

zhiyuanliang-ms:
#333


linglingye001 (approved):

src/appConfigurationImpl.ts
```diff
@@ -659,7 +659,7 @@ export class AzureAppConfigurationImpl implements AzureAppConfiguration {
 
                 const watcher: SettingWatcher = this.#sentinels.get(watchedSetting)!; // watcher should always exist for sentinels
                 const isDeleted = response === undefined && watcher.etag !== undefined; // previously existed, now deleted
-                const isChanged = response && response.statusCode === 200 && watcher.etag !== response.etag; // etag changed
+                const isChanged = response && Number(response.statusCode) === 200 && watcher.etag !== response.etag; // etag changed
```
linglingye001, line 662: Better to have comment in the code/helper

PR #341, opened 2026-07-23
load feature flags from new endpoint

linglingye001:
**Summary**
Adds support for loading feature flags from new endpoint (via `FeatureFlagClient.listFeatureFlags`), in addition to the classic key-value path. When feature flags are enabled, the provider now loads flags from both sources and merges them, with flags from the new endpoint taking precedence over classic ones of the same name. Reference https://github.com/Azure/AppConfiguration-DotnetProvider/pull/738

**Key changes**

- Added `appConfigClient.ts`: an `IAppConfigurationClient` interface plus `AppConfigClient` implementation that wraps both `AppConfigurationClient` (classic KV) and `FeatureFlagClient` (new FF endpoint). Each method applies request tracing before delegating to the underlying SDK client.

- Added `featureFlagConverter.ts` with `convertToMicrosoftSchema()`, which maps the SDK's typed `FeatureFlag` (camelCase) into the Microsoft Feature Management schema consumed by `feature_management.feature_flags`.

- Load and merge on init and on refresh: `#loadClassicFeatureFlags()` + `#loadFeatureFlags(`) → `#setFeatureFlags()` dedups by name (new endpoint supersedes classic).

- Change detection covers both paths


src/featureManagement/featureFlagConverter.ts
zhiyuanliang-ms, line 1: `FeatureFlagConditions`, `FeatureFilter`, .... all of these types/interfaces/enum are exported by the Azure SDK. We should use them instead of abusing `any` type.

src/requestTracing/utils.ts
```diff
@@ -58,64 +50,17 @@ export interface RequestTracingOptions {
...
-
-function applyRequestTracing<T extends OperationOptions>(requestTracingOptions: RequestTracingOptions, operationOptions?: T) {
+export function applyRequestTracing<T extends OperationOptions>(requestTracingOptions: RequestTracingOptions, operationOptions?: T) {
     const actualOptions = { ...operationOptions };
     if (requestTracingOptions.enabled) {
         actualOptions.requestOptions = {
             ...actualOptions.requestOptions,
-            customHeaders: {
-                ...actualOptions.requestOptions?.customHeaders,
+            headers: {
```
zhiyuanliang-ms, line 62: Is this change intentional? Or it is accidently introduced by ai?

It should be [`customHeaders`](https://github.com/Azure/azure-sdk-for-js/blob/main/sdk/core/core-client/src/interfaces.ts#L132)

test/requestTracing.test.ts
```diff
@@ -246,7 +246,7 @@ describe("request tracing", function () {
     it("should have filter type in correlation-context header if feature flags use feature filters", async () => {
         let correlationContext: string = "";
         const listKvCallback = (listOptions) => {
-            correlationContext = listOptions?.requestOptions?.customHeaders[CORRELATION_CONTEXT_HEADER_NAME] ?? "";
+            correlationContext = listOptions?.requestOptions?.headers?.[CORRELATION_CONTEXT_HEADER_NAME] ?? "";
```
zhiyuanliang-ms: I am thinking maybe we should verify the request header instead of checking requestOptions in the request tracing test.

src/appConfigClient.ts
```diff
@@ -0,0 +1,100 @@
...
+        tracingOptions: RequestTracingOptions
+    ): ReturnType<FeatureFlagClient["listFeatureFlags"]>;
+}
+
+/**
+ * The default @see IAppConfigurationClient implementation, backed by the @see AppConfigurationClient
+ * and @see FeatureFlagClient SDK clients. Request tracing is applied per call before delegating to the
+ * corresponding SDK client.
+ */
+export class AppConfigClient implements IAppConfigurationClient {
```
zhiyuanliang-ms: I think we can name this class (and file) as `AppConfigurationClient` and then 
```ts
import {
  AppConfigurationClient as ConfigurationClient,
} from "@azure/app-configuration";
```

src/configurationClientManager.ts
```diff
@@ -1,7 +1,8 @@
 // Copyright (c) Microsoft Corporation.
 // Licensed under the MIT license.
 
-import { AppConfigurationClient, AppConfigurationClientOptions } from "@azure/app-configuration";
+import { AppConfigurationClient, AppConfigurationClientOptions, FeatureFlagClient } from "@azure/app-configuration";
```
zhiyuanliang-ms:
```ts
export interface FeatureFlagClientOptions extends AppConfigurationClientOptions {}
```
It is technically feasible to use `AppConfigurationClientOptions` for `FeatureFlagClient`.
Do we intentionally use the same options for both clients?
If not, we should have separated `FeatureFlagClientOptions`

src/appConfigClient.ts
```diff
@@ -0,0 +1,100 @@
...
+    ListConfigurationSettingsOptions,
+    ListFeatureFlagsOptions
+} from "@azure/app-configuration";
+import { RequestTracingOptions, applyRequestTracing } from "./requestTracing/utils.js";
+
+/**
+ * A single client abstraction for an Azure App Configuration endpoint that exposes the subset of
+ * operations the provider needs from both the @see AppConfigurationClient and the @see FeatureFlagClient.
+ */
+export interface IAppConfigurationClient {
```
zhiyuanliang-ms: We don't need the `IAppConfigurationClient` abstraction in javascript. `appConfigurationImpl` should use the `AppConfigClient` directly. In JS, when you define an interface, in most of cases, the motivation is to export it for public use. But the AppConfigurationClient is completely for internal usage.

test/utils/testHelper.ts
```diff
@@ -217,6 +312,11 @@ function mockAppConfigurationClientListConfigurationSettings(pages: Configuratio
 
         return getMockedHeadIterator(pages, listOptions);
     });
+
+    // The dedicated feature flag endpoint is queried in addition to feature flags; default to an empty result set.
+    sinon.stub(FeatureFlagClient.prototype, "listFeatureFlags").callsFake((listOptions) => {
```
zhiyuanliang-ms: Do we have similar method like `checkConfigurationSettings` to send `HEAD` request for feature flags?

zhiyuanliang-ms: I'd like to have it in a different function or update the function name. The current function name is related to configuration settings.

test/utils/testHelper.ts
```diff
@@ -249,6 +354,10 @@ function mockAppConfigurationClientLoadBalanceMode(pages: ConfigurationSetting[]
         countObject.count += 1;
         return getMockedHeadIterator(pages, listOptions);
     });
+    sinon.stub(clientWrapper.client, "listFeatureFlags").callsFake((listOptions) => {
+        countObject.count += 1;
```
zhiyuanliang-ms: I think we should update the countObject to record call numbers for both feature flag and configuration settings.
e.g. kvCallCount and ffCallCount (let copilot come up with a better name)

test/utils/testHelper.ts
```diff
@@ -431,6 +546,15 @@ const createMockedSnapshotReference = (key: string, snapshotName: string): Confi
     isReadOnly: false,
 });
 
+// Creates an enhanced feature flag as returned by the dedicated feature flag endpoint (FeatureFlagClient.listFeatureFlags).
+const createMockedEnhancedFeatureFlag = (name: string, props?: any) => Object.assign({
```
zhiyuanliang-ms, line 562: How about enhancing the existing `createMockedFeatureFlag` by passing a flag `isConfigurationSetting`

If `false`, it will create a `FeatureFlag` shape object

test/afd.test.ts
```diff
@@ -97,7 +97,7 @@ describe("loadFromAzureFrontDoor", function() {
             userAgent = headerPolicy.headers.get("User-Agent");
         }
 
-        expect(userAgent).satisfy((ua: string) => ua.startsWith("javascript-appconfiguration-provider"));
+        expect(userAgent).satisfy((ua: string) => ua.includes("javascript-appconfiguration-provider"));
```
zhiyuanliang-ms, line 100: What's the reason of this change? Is there any behavior change on how we set user agent?

src/featureManagement/featureFlagConverter.ts
```diff
@@ -0,0 +1,87 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT license.
+
+import { FeatureFlag } from "@azure/app-configuration";
+
+/**
+ * Converts @see FeatureFlag returned by the feature flag endpoint into the
+ * Microsoft Feature Management schema object (snake_case) used within the `feature_management.feature_flags`
```
zhiyuanliang-ms:
```suggestion
 * Microsoft Feature Flag schema object used within the `feature_management.feature_flags`
```

src/featureManagement/featureFlagConverter.ts
```diff
@@ -0,0 +1,87 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT license.
+
+import { FeatureFlag } from "@azure/app-configuration";
+
+/**
+ * Converts @see FeatureFlag returned by the feature flag endpoint into the
```
zhiyuanliang-ms:
```suggestion
 * Converts @see FeatureFlag into the
```

src/featureManagement/featureFlagConverter.ts
```diff
@@ -0,0 +1,87 @@
...
+ * Microsoft Feature Management schema object (snake_case) used within the `feature_management.feature_flags`
+ * array. This mirrors the shape produced by parsing a classic feature flag key-value, so that downstream
+ * feature management parsing and the provider's telemetry/tracing logic are unchanged.
+ */
+export function convertToMicrosoftSchema(featureFlag: FeatureFlag): any {
```
zhiyuanliang-ms:
```suggestion
export function convert(featureFlag: FeatureFlag): any {
```

src/appConfigurationImpl.ts
```diff
@@ -769,10 +817,64 @@ export class AzureAppConfigurationImpl implements AzureAppConfiguration {
...
+        const isChanged = await this.#executeWithFailoverPolicy(funcToExecute);
+        return isChanged;
+    }
+
+    /**
+     * Checks whether the enhanced feature flag collection from the dedicated feature flag endpoint has changed.
+     * @param selectors - The @see PagedSettingsWatcher of the enhanced feature flag collection.
+     * @returns true if the enhanced feature flag collection has changed, false otherwise.
+     */
+    async #checkEnhancedFeatureFlagsChange(selectors: PagedSettingsWatcher[]): Promise<boolean> {
```
zhiyuanliang-ms, line 861: Is `HEAD` request supported for new feature flag?

test/afd.test.ts
linglingye001: behavior change is from SDK side, I think we have discussed this offline before. Now the useragent is like this ` "User-Agent": "azsdk-js-app-configuration/1.13.0-beta.1 javascript-appconfiguration-provider/2.5.0-preview azsdk-js-app-configuration/1.13.0-beta.1 core-rest-pipeline/1.25.0 Node/22.19.0 (Windows_NT 10.0.26200; x64)",`

src/requestTracing/utils.ts
linglingye001: New version SDK introduced `"@azure-rest/core-client": "^2.5.0"` and we use https://github.com/Azure/azure-sdk-for-js/blob/main/sdk/core/core-client-rest/src/common.ts#L161 now.

zhiyuanliang-ms: Fixed in https://github.com/Azure/AppConfiguration-JavaScriptProvider/pull/348

src/appConfigurationImpl.ts
```diff
@@ -628,25 +614,85 @@ export class AzureAppConfigurationImpl implements AzureAppConfiguration {
         }
     }
 
-    /**
-     * Loads feature flags from App Configuration to the local configuration.
-     */
-    async #loadFeatureFlags() {
-        const loadFeatureFlag = true;
-        const featureFlagSettings: ConfigurationSetting[] = await this.#loadConfigurationSettings(loadFeatureFlag);
+    async #loadFeatureFlags(): Promise<ConfigurationSetting[]> {
```
zhiyuanliang-ms, line 617: loadFeatureFlags still returns void to make it symmetric to loadSelectedKeyValues

This will make the #initializeWithRetryPolicy code path very clear (reader don't need to know about the context about old/new feature flag business logic)


All the business logic of loading and merging old and new feature flags happen in loadFeatureFlags

We can two new methods:
1. loadFeatureFlagConfigurationSettings returns ConfigurationSettings[]
This method should reuse the previous loadConfigurationSettings(boolean)

Currently, #loadSelectedKeyValues and #loadFeatureFlags share a lot of similar code.
But if we keep using the previous loadConfigurationSettings(isLoadFeatureFlag), we can save duplicated code.

2. loadEnhancedFeatureFlags returns FeatureFlag[]

loadFeatureFlags should be 

```
ffs =  loadFeatureFlagConfigurationSettings();
enhancedFfs = loadEnhancedFeatureFlags();
// remaining logic of merging ffs and enhanced ffs
```

src/appConfigurationImpl.ts
```diff
@@ -1223,25 +1357,27 @@ function areTagFiltersEqual(tagsA?: string[], tagsB?: string[]): boolean {
...
-    selectors.forEach(selector => {
+    // Deep clone so the caller's option objects are never mutated.
+    const clonedSelectors = structuredClone(selectors);
+    clonedSelectors.forEach(selector => {
         if (selector.keyFilter) {
             selector.keyFilter = `${featureFlagPrefix}${selector.keyFilter}`;
         }
     });
-    return getValidSettingSelectors(selectors);
+    return getValidSettingSelectors(clonedSelectors);
```
zhiyuanliang-ms:
```suggestion
    // Create prefixed copies because the original selectors are also used as unprefixed selectors for enhanced feature flags.
    const prefixedSelectors = selectors.map(selector => ({
        ...selector,
        keyFilter: selector.keyFilter
            ? `${featureFlagPrefix}${selector.keyFilter}`
             : selector.keyFilter
    }));
    return getValidSettingSelectors(prefixedSelectors);
```

src/appConfigurationImpl.ts
```diff
@@ -628,25 +614,85 @@ export class AzureAppConfigurationImpl implements AzureAppConfiguration {
...
         }
 
-        // parse feature flags
-        const featureFlags = await Promise.all(
-            featureFlagSettings.map(setting => this.#parseFeatureFlag(setting))
+        // Track whether enhanced feature flags were loaded.
+        this.#useEnhancedFeatureFlag = enhancedFeatureFlags.length > 0;
+
+        // Exclude any feature flags that are superseded by an enhanced feature flag with the same name.
+        const ineligibleFeatureFlagKeys = new Set(enhancedFeatureFlags.map(ff => featureFlagPrefix + ff.name));
```
zhiyuanliang-ms, line 684:
```suggestion
        const supersededFeatureFlagKeys = new Set(enhancedFeatureFlags.map(ff => featureFlagPrefix + ff.name));
```

src/appConfigurationImpl.ts
```diff
@@ -989,31 +1106,48 @@ export class AzureAppConfigurationImpl implements AzureAppConfiguration {
         }
         const featureFlag = JSON.parse(rawFlag);
```
zhiyuanliang-ms:
```suggestion
        const parsedFeatureFlag = JSON.parse(rawFlag);
```

zhiyuanliang-ms: Make it symmetric to another parseFf function

src/featureManagement/featureFlagConverter.ts
linglingye001: FeatureFlag related types exported in SDK are different with those defined by MS schema, that's why we need the converter. If imported them from SDK we still need to define the corresponding structs with snake_case field names. What concerns do you have about using the `any` type?
```javascript
/** The conditions that must be met for the feature flag to be enabled. */
export interface FeatureFlagConditions {
  /** The requirement type for the conditions. */
  requirementType?: RequirementType;
  /** The filters that will conditionally enable or disable the flag. */
  filters?: FeatureFilter[];
}
/** A filter that conditionally enables or disables a feature flag. */
export interface FeatureFilter {
  /** The name of the filter. */
  name: string;
  /** The parameters used by the filter. */
  parameters?: { [propertyName: string]: string };
}
```

src/appConfigurationImpl.ts
linglingye001: I understand your idea for using `loadFeatureFlags` to handle all feature flag related logic. Currently, across different languages, the term feature flag refers to the classic feature flags, and the existing implementation is for consistency.

From the naming perspective, `loadFeatureFlags` and `loadEnhancedFeatureFlags` look symmetrical. However, behavior-wise, `loadFeatureFlags` also loads enhanced feature flags, which feels odd to me.

src/appConfigurationImpl.ts
linglingye001: this is consistent with dotnet provider https://github.com/Azure/AppConfiguration-DotnetProvider/pull/738/changes#diff-307c2ab55df3e48a953dc97f7755b5dcb6a20849b8ceda5212cdc007152c9181R406

test/utils/testHelper.ts
linglingye001: In this PR, "feature flag" specifically refers to the classic feature flag, while "enhanced feature flag" refers to the new FF. We intentionally keep the handling of classic and enhanced feature flags separate, including their loading logic, which is consistent with the .NET provider.

Our tests currently use `createMockedFeatureFlag` and `createMockedEnhancedFeatureFlag` respectively for these two scenarios. Given that classic and enhanced feature flags are distinct object types returned by different clients, separating their handling makes the intent clearer.

If `createMockedFeatureFlag` were used for both scenarios, the term "feature flag" would implicitly refer to both classic and enhanced feature flags, which blurs the distinction between the two concepts and may introduce  confusion.

src/appConfigurationImpl.ts
```diff
@@ -769,10 +817,64 @@ export class AzureAppConfigurationImpl implements AzureAppConfiguration {
...
+                for await (const page of pageIterator) {
+                    const serverResponseTime: Date = this.#getMsDateHeader(page);
+                    if (i >= pageWatchers.length) {
+                        return true;
+                    }
+
+                    const lastServerResponseTime = pageWatchers[i].lastServerResponseTime;
+                    let isResponseFresh = false;
+                    if (lastServerResponseTime !== undefined) {
+                        isResponseFresh = serverResponseTime > lastServerResponseTime;
```
zhiyuanliang-ms, line 835: In .NET, we use ">=". But I guess this behavior difference has existed before this PR. Let's align with .NET even though this difference will not affect anything in reality

src/appConfigurationImpl.ts
```diff
@@ -769,10 +817,64 @@ export class AzureAppConfigurationImpl implements AzureAppConfiguration {
...
+                    }
+                    if (isResponseFresh &&
+                        getStatusCode(page._response.status) === 200 && // conditional request returns 304 if not changed
+                        page.etag !== pageWatchers[i].etag) {
+                        return true;
+                    }
+                    i++;
+                }
+            }
```
zhiyuanliang-ms, line 848: In .NET provider, we have an [additional check](https://github.com/Azure/AppConfiguration-DotnetProvider/blob/main/src/Microsoft.Extensions.Configuration.AzureAppConfiguration/Extensions/ConfigurationClientExtensions.cs#L108) to see whether there is more old pages. 

To align with .NET, we should add
```js
if (i < pageWatchers.length) {
    return true;
}
```
after the for loop here.

But this behavior difference has existed before this PR and it will not affect anything in reality, because we will have an empty page after a full page. That empty page will have different etag.
But add this additional check can make code more robust.

src/featureManagement/featureFlagConverter.ts
linglingye001: Discussed offline, replaced the untyped output with MS FM schema types.

src/featureManagement/featureFlagConverter.ts
```diff
@@ -0,0 +1,92 @@
...
+export function convert(featureFlag: AzAppConfigFeatureFlag): FeatureFlag {
+    const result: FeatureFlag = {
+        id: featureFlag.name,
+        enabled: featureFlag.enabled,
+        conditions: {
+            client_filters: (featureFlag.conditions?.filters ?? []).map(filter => {
+                const clientFilter: FeatureFilter = { name: filter.name };
+                if (filter.parameters != null) {
+                    clientFilter.parameters = Object.fromEntries(
+                        Object.entries(filter.parameters).map(([name, value]): [string, unknown] => [name, JSON.parse(value) as unknown]));
```
zhiyuanliang-ms: This line cannot handle the case that parameter value is not a valid json.

For example:

```js
const parameters = {
  Start: "Mon, 01 Jan 2024 00:00:00 GMT"
};

const converted = Object.fromEntries(
  Object.entries(parameters).map(([name, value]) => [name, JSON.parse(value)])
);

console.log(converted);
```
It will throw `SyntaxError: Unexpected token 'M', "Mon, 01 Ja"... is not valid JSON`

zhiyuanliang-ms: In .NET PR: We only parse json when the parameter value seems to be json value

Image

src/requestTracing/utils.ts
zhiyuanliang-ms, line 143: We missed use enhanced feature flag here.

src/featureManagement/featureFlagConverter.ts
```diff
@@ -0,0 +1,92 @@
...
+    if (featureFlag.conditions?.requirementType != null) {
+        result.conditions.requirement_type = featureFlag.conditions.requirementType;
+    }
+
+    // variants: value -> configuration_value, statusOverride -> status_override
+    if (featureFlag.variants != null) {
+        result.variants = featureFlag.variants.map(variant => {
+            const resultVariant: FeatureFlagVariant = { name: variant.name };
+            if (variant.value !== undefined) {
+                resultVariant.configuration_value = variant.value;
```
zhiyuanliang-ms: We didn't respect the variant content type here.

Image

In .NET PR, we will parse json value:

Image

linglingye001: updated to keep consistent with .NET provider

zhiyuanliang-ms (approved):

src/featureManagement/featureFlagConverter.ts
```diff
@@ -0,0 +1,117 @@
...
+        if (featureFlag.telemetry.metadata != null) {
+            telemetry.metadata = featureFlag.telemetry.metadata;
+        }
+        result.telemetry = telemetry;
+    }
+
+    return result;
+}
+
+function parseParameterValue(value: string): unknown {
```
zhiyuanliang-ms: The parameter passed in will be `unknown` type

src/featureManagement/featureFlags.ts
```diff
@@ -0,0 +1,54 @@
...
+    FeatureFlagTelemetryConfiguration as AzAppConfigFeatureFlagTelemetryConfiguration,
+    FeatureFlagVariantDefinition as AzAppConfigFeatureFlagVariantDefinition,
+    GroupAllocation as AzAppConfigGroupAllocation,
+    PercentileAllocation as AzAppConfigPercentileAllocation,
+    UserAllocation as AzAppConfigUserAllocation
+} from "@azure/app-configuration";
+
+export interface FeatureFilter {
+    name: AzAppConfigFeatureFilter["name"];
+    parameters?: Record<string, unknown>;
```
zhiyuanliang-ms:
```suggestion
    parameters?: Record<string, any>;
```

src/featureManagement/featureFlagConverter.ts
```diff
@@ -0,0 +1,117 @@
...
+    const trimmedValue = value.trim();
```
zhiyuanliang-ms, line 105: We need to add check for `value` can be `undefined` or `null`.

src/appConfigurationClient.ts
```diff
@@ -0,0 +1,91 @@
...
+    }
+}
+
+export function getFeatureFlagClientOptions(options?: ConfigurationClientOptions): FeatureFlagClientOptions | undefined {
+    if (options === undefined) {
+        return undefined;
+    }
+
+    return {
+        ...options,
```
zhiyuanliang-ms, line 89: We should not allow user to override [apiversion](https://github.com/Azure/azure-sdk-for-js/blob/main/sdk/appconfiguration/app-configuration/src/models.ts#L98) for feature flag client.

src/appConfigurationClient.ts
```diff
@@ -0,0 +1,91 @@
...
+    constructor(endpoint: string, credential: TokenCredential, options?: ConfigurationClientOptions);
+    constructor(
+        connectionStringOrEndpoint: string,
+        credentialOrOptions?: TokenCredential | ConfigurationClientOptions,
+        options?: ConfigurationClientOptions
+    ) {
+        const credentialPassed = instanceOfTokenCredential(credentialOrOptions);
+        const configurationClientOptions = credentialPassed
+            ? options
+            : credentialOrOptions as ConfigurationClientOptions | undefined;
```
zhiyuanliang-ms:
```suggestion
        const configurationClientOptions = credentialPassed ? options : credentialOrOptions as ConfigurationClientOptions | undefined;
```

